PR #2, opened 2019-03-07
Bugfix/melding view

mehran-naghizadeh:
This PR looks forward to addressing some issues in the UI. When users want to take a look at a toezicht, they are provided with things that they don't expect: 

- Warnings about mandatory form fields

- Buttons for deleting urls associated with the toezicht 

- The chance to edit urls

The way a toezicht is presented for viewing is via the same route as is used for editing. Here, a new route is created to separate the two tasks. Templates associated with these two routes, both rely on the same component but now each route can pass its desireable argument to the template. The most important argument for us is editMode which tells viewing mode apart from editing mode. The template can easily reshape itself according to this argument.


madnificent:
I did not run the code, merely checked the style.  If it is fully functional I will run it and check the behaviour.  I expect you will do the same :)


# Use spaces instead of tabs

Please use spaces instead of tabs in the code.  It makes the code consistent regardless of the indentation offset configured in the editor.  You should be able to update your editor's preferences.

# Model hook in inzendingen/view route

You don't need the parens around `await` in the following code-sample.  We could also make the indentation somewhat nicer.

``` javascript
  async model(params) {
    let models =  (await this.store.query('form-solution',
                                                 {'filter[inzending-voor-toezicht][id]': params.id,
                                                  include: ['inzending-voor-toezicht.status',
                                                            'inzending-voor-toezicht.last-modifier',
                                                            'inzending-voor-toezicht.bestuurseenheid',
                                                            'inzending-voor-toezicht.inzending-type',
                                                            'inzending-voor-toezicht.besluit-type',
                                                            'inzending-voor-toezicht.bestuursorgaan',
                                                            'inzending-voor-toezicht.tax-rates',
                                                            'inzending-voor-toezicht.file-addresses'
                                                           ].join(',')}));
    return models.firstObject;
  }

```

Could be transformed to

``` javascript
  async model(params) {
    let models =  await this.store.query(
      'form-solution',
      {'filter[inzending-voor-toezicht][id]': params.id,
       include: ['inzending-voor-toezicht.status',
                 'inzending-voor-toezicht.last-modifier',
                 'inzending-voor-toezicht.bestuurseenheid',
                 'inzending-voor-toezicht.inzending-type',
                 'inzending-voor-toezicht.besluit-type',
                 'inzending-voor-toezicht.bestuursorgaan',
                 'inzending-voor-toezicht.tax-rates',
                 'inzending-voor-toezicht.file-addresses'
                ].join(',')});
    return models.firstObject;
  }
```

which is arguably easier to read.  Not sure if there's a better way of putting things.


# Use of isNew

The definition of `isNew` doesn't seem to be in the code but I might be missing it.  In any case, as discussed over chat, perhaps we should use `(not isSent)` instead.

# Block with await in inzending-edit component

Following block seems strange to me

``` javascript
  {{#if (await (or isNew editMode fileAddresses.length))}}
    <div class="form__row u-spacer">
      {{toezicht/url-box urls=fileAddresses disabled=isSent onDelete=(action "deleteFileAddress")}}
    </div>
  {{/if}}
```

It's not clear to me what the intention of `(await (or isNew editMode fileAddresses.length))` is.  It should supposedly become `(or isNew editMode (await fileAddresses.length))` but I strongly doubt you need the await then, and it's not clear to me why we are checking for all these options.  There's likely a very reasonable answer to this, it' just that I don't know it yet :-)

# Link to inzending and value of new route

The link to inzending in the inzendingen-table always renders the view route.  IIRC we should be able to update the inzending when it's been saved but not sent.  That means the edit route should be used in the case the file has not been sent yet.

Introducing a new route makes sense from a distance.  With the current code I wonder if it simplifies things.  The component handles all logic and now receives a parameter `editMode` to see if we are in edit mode.  However, we are always in edit mode when the document has not been sent yet.  The introduction of the new route allows people to press the back button and suddenly land in the edit mode again (unless the document was sent, because we have the two checks in most places in the code).

It seems to make sense to remove the new route and to enable editing always when the document has not been sent yet.  Note that I derive this without running the code, perhaps I'm missing some things.

mehran-naghizadeh:
Explaining (or isNew editMode fileAddresses.length)

This statement checks three things:
	1) isNew -> indicates creation mode: this was already there in the model, I didn’t create it. In the creation mode we definitely need to render the url-box component
	2) editMode -> indicates editing mode: This argument is passed to the component by the calling template(/route).  Again, in the edit mode we need to render the url-box component
	3) fileAddresses.length -> This may need some explanation. When the other two arguments are false, it means that we are in the view mode. Now, there is the possibility that the toezicht under review doesn’t contain any fileAddress. Here, we check this. If there is nothing in the array, we don’t need to render the component.

I hope this is clear. Though, there might be better ways for doing this.

mehran-naghizadeh:
Why I created a new route

When users click on ‘Bekijk’ buttons, they are transitioned to a new route, the call is `{{#link-to 'toezicht.inzendingen.view' row.id}}` I needed away to tell this route that we are just viewing, don’t let users to edit or change anything. It would be nice if I could pass a parameter to the route, but unfortunately, I couldn’t find a way to do so. So, I thought of another option: making a new route.

Of course, another option is to let the component itself decide about its working mode. For that, I could check if the toezicht is verstuurd or not. But, I have learned from Felix that a component is meant to be an atom. I may have made a mistake to think that letting the component decide about its working mode will break this rule.

So, I am not defending the new route, it was just the solution I could find.

madnificent:
Yes, that makes sense.  It seems the component has roughly everything it needs in order to derive this knowledge so I thought it made sense.  Perhaps that's not ideal either.

From a distance, it seems we care about two things: are we in edit mode (anything before save) and are we new (to possibly render a different header/footer).  Perhaps we can take this opportunity to refactor the code so it's easier to understand.  It's not the shortest task possible, but we'll learn quite a bit.

Whether the form is editable (meaning it's not been sent yet) is indeed something a consuming entity could pass.  ie: it could be a computed property on a controller.  I suggest introducing an `editable` variable in the component which defaults to false, and supplying that to the component.  I could extend the `root-node` dependency so it would also accept an `editable` argument, rather than only a `show`.  Until then, it would be `show=(not editable)`.  This would likely make the code clearer to understand.  It would give us a single variable to depend on for the conditional template.  We wouldn't need a custom route to determine whether this is editable or not, the controller could have an alias to `model.inzendingVoorToezicht.status.isVerstuurd`.

The `isNew` is very similar, yet slightly different.  It basically means "if the model hasn't been saved before".  This either has to be supplied to the component, or it could be a computed property on the model which is to be saved.  I'm not sure what the best behaviour would be.  Although I'd traditionally put this logic outside of the component, we basically have all knowledge we need: `model.isNew`.  Guess this is ripe for further discussion as we make our minds up on this point.  Let's keep it as an argument for now, but specify a default value of false in the component.

Would you like to continue and refactor this code so it's a more clear from a consumer and a producer's perspective?

mehran-naghizadeh:
Yes, of course.

madnificent:
From a functionality point of view, the code seems to be largely ok.  I did not test it thoroughly, but a quick case seemed to make things behave roughly the way they should.  It is certainly usable.

# Movable logic in inzending-edit component and url-box

The logic around this component is confusing.  It's not readily clear why this logic is there.  If you think about it for a few minutes, it makes sense, but when you see it from a distance it's not obvious.

``` handlebars
  {{#if (or (not isSent) fileAddresses.length )}}
```

I would suggest moving this whole logic into the `url-box` component.  I don't think that makes it very clear either, but at least we'll be able to abstract the logic there and guide the reader in our intentions.

# url-line

Calling the url-line in `url-box` shows something like:

``` handlebars
      {{#if saveAddress.isRunning}}
        Link aan het controleren...
      {{else}}
        {{#each urls as |url|}}
          <tr>
            {{toezicht/url-line url=url editable=(not disabled) onDelete=onDelete}}
          </tr>
        {{/each}}
      {{/if}}
```

Although it's readable, we can make this cleaner:

## Property saveAddress.isRunning

I don't think we have logic for the following bit in place:

``` handlebars
      {{#if saveAddress.isRunning}}
        Link aan het controleren...
      {{else}}
```

The `saveAddress` task is still there, there is a `save` action, but it is never called.  I suggest removing this task, and the save action that belongs to it for now.  Because we currently only fetch the files when the form is submitted, there's no need for the save.  We've learned from it and we should take it with us.  Perhaps we'll introduce the logic again later.

Side remark: I'm not sure you're allowed to put a sentence in a tbody, rather than table cells.

If we keep the logic with commented code, we will keep wandering about the feature.  It's sufficiently simple to implement that you'll be able to implement it again in a much shorter time.

This would have an impact both in the template, as in the url-box component.

## Move TR into url-line

The `url-line` component is responsible for the `td` tags inside of its template.  Hence it knows that it's rendering in a table.  We might as well move this `tr` inside of there too.

We currently have

``` handlebars
        {{#each urls as |url|}}
          <tr>
            {{toezicht/url-line url=url editable=(not disabled) onDelete=onDelete}}
          </tr>
        {{/each}}
```

Which would then be show as:

``` handlebars
        {{#each urls as |url|}}
          {{toezicht/url-line url=url editable=(not disabled) onDelete=onDelete}}
        {{/each}}
```

It's not a big deal, but it looks a bit easier on the eyes.  If we introduce the component, we might as well let it handle the whole thing.

## Wrapping tag in url-line component

The url-line component has a wrapping `div` tag because we don't force it not to have one.  I doubt we're technically allowed to put div tags somewhere between `tbody > tr > td`.  Inside of the `td` that would be fine though.

You can do this by setting `tagName: ""` in the `url-line` component.

## Logic for editable an non-editable url-line

The rendering of the input and link for the url-line seems duplicated.

``` handlebars
  {{#if editable}}
    {{input
      class="input-field input-field--block input-field--small"
      placeholder="http://wwww.uw-bestuur.be/specifiek-document"
      name="URL-besluit-URL"
      value=url.address
    }}
  {{else}}
    <a href={{url.address}} target="_blank">{{input
      class="input-field input-field--block input-field--small"
      placeholder="http://wwww.uw-bestuur.be/specifiek-document"
      name="URL-besluit-URL"
      value=url.address
      disabled=true
    }}</a>
```

I'm not sure what the right way would be to handle this case as I don't have a clean way to remove the duplication.  This may be the best we have.  Luckly for me it also doesn't seem to behave the way we'd expect.  You should try to press the link, it doesn't open up a page.  At least not in Firefox.  As such, I suggest we remove some of the features, and then see if @mietcls can design a nice link for us.  From then on, we'll revisit.

From there on, I suggest to only render the input (not the anchor tag).  The only difference between these two should be the `disabled` property, which we can bind correctly using our `editable` property.

It makes sense to keep track of the feature in which we can click somewhere to open the original URL, yet it's not the most important.

## Text for download link

The download of our archived document is the following:

``` handlebars
    <a href="{{url.replicatedFile.downloadLink}}" download><p>Download voorbeeld</p></a>
```

The download link says `Download voorbeeld`.  This translates to "Download example".  We're downloading an archived document here, so I'd rephrase this to `Download archiefbestand`.

The wrapping `<p>` tag is likely redundant, so we can remove that too.


# use of editMode

Is the editMode still in use in this application?  I see it in `toezicht/inzendingen/edit.hbs` and there is a default value in `inzending-edit`.  Assuming it is not used anywhere, we should remove it so our code stays clean.  People will assume it does something.


# url-box tests

The url-box tests are being removed in this PR.  This is likely my fault.  If you want to keep them, then commit them again :-)

mehran-naghizadeh:
Thanks for the comments. I keep the wrapping <p> tag in the next push because I have replaced the {{input}} helper with an <a> tag. Take a look at it (in the browser) and if you didn't like it, I will remove both tags. The tests (url-line, url-box) are intentionally kept out. My Intention is to focus on one thing then go to the next. In fact, having 3 parallel branches had made a lot of confusion.

madnificent:
Agree!  This has become broad enough as it is.  I'll take a peek.

madnificent:
This is starting to look very nice.  We're cleaning up things which are partially further than the original intent. :-)

# url-box disabled

I see we use the `disabled` option.  It has not explicitly received a default value in the component.  I'd add that so people know it's an option.

# inzending-edit

Your take on defining the needsUrlBox property is nice.  It belongs there more than somewhere else indeed.  We should, not in this PR, haul over the inzending-edit component.  It's becoming quite large making it hard to understand what is going on exactly.  Good idea!

The documentation seems to contain a few typos.  I've corrected them in the snippet below.  I've introduced two variables in the computed property which might make the code easier to read.  Don't feel obliged to use that part of the snippet, it may not be better.

``` javascript
  /**
   * url-box is used to either show existing urls or enter new ones
   * It is needed in two cases
   *  1. The inzending is not yet finalized/sent: !isSent
   *  2. The inzending has been sent and it contains at least one url: fileAddress.length > 0
   */
  needsUrlBox: computed('isSent', 'fileAddresses.length', function() {
    const notSent = !this.isSent;
    const hasFileAddresses = this.get('fileAddresses.length') > 0;
    return notSent || hasFileAddresses;
  }),
```

# url-line archiefbestand

We have an `{{else}}` for the download link.  I'd drop the else block and we'll ask @mietcls if we should have a message for this clause or not.  It will be confusing if the file is not there as users will likely not really know what it really means.  It currently also appears when the user is entering a new URL, so it feels strange.  From a technical perspective it makes much sense, and them being able to download the files to verify what we have discovered is important, but it's weird for those that have no clue how the system works.  Most users will not check our GitHub or technical architecture.  I'd drop the else clause until there is clarity on how we will communicate this to users.

# url-line original location download

From a technical point of view it makes sense to convert their input into a link.  But now suddenly there are two links in that box.  One with an archive document and one with their site in it.  This doesn't give the user the feedback that this is the input they entered.  My reason for going with the input was to ensure that this would give the visual feel that they entered that data.  The link portrays a different thing.

# editMode in inzendingen/edit template

When we render the `toezicht/inzending-edit` component, we still supply the `editMode` property.

``` handlebars
{{toezicht/inzending-edit model=model editMode=true}}
```

This doesn't exist anymore and can be dropped.

mehran-naghizadeh:
> My reason for going with the input was to ensure that this would give the visual feel that they entered that data

I completely agree with you on this. Having two visually similar links side by side is misleading and confusing. We definitely need to render one of them differently. This suggest keeping the original url as it used to be: an {{input}} field. The problem here is that when you set {{input ... disabled=true}}, the input field's background becomes dark gray while its text remains black. This combination can't be easily distinguished from the gray background.
On the other hand, having something to click on is most probably what users will be expecting. After all, there is a call-to-action associated with any url: click on me! This complication can be easily solved by a good visual design of our wise designer.

madnificent:
I agree on the input front.  It's similar to the other inputs though, so it seems reasonable to assume the design will cope with this if that's necessary.  I notice we can't select the text when the input is set to disabled.  Can be discussed later.

There's some things that could be done to make this code more readable but they're really not related to this feature.

PR #6, opened 2019-04-17
Feature/toezicht form fields

mehran-naghizadeh:
This PR seeks to replace local toezicht components with components from the **@lblod/ember-toezicht-form-fields addon**. These include url-box and url-line which are _specific to the toezicht_ and all input-fields components which can also be used elsewhere. The addon is available at 
https://github.com/lblod/ember-toezicht-form-fields


package.json
```diff
@@ -75,10 +76,10 @@
     "eslint-plugin-ember": "^5.2.0",
     "loader.js": "^4.7.0",
     "postscribe": "^2.0.8",
-    "toezicht-form-fields": "x.x.x",
```
cecemel: Fyi: I thought only a '*' would work. Don't think "x.x.x.x" works.
Furthermore try to avoid committing such things

PR #7, opened 2019-05-14
Feature/leidinggevenden

mehran-naghizadeh:
[Don't merge]
This PR adds a new route (with all its sub routes) to the locket frontend app: **leidinggevendenbeheer**


erikap:
@mehran Can you try to resolve the linting errors that appear in the console when starting the application?

mehran-naghizadeh:
> @mehran-naghizadeh Can you try to resolve the linting errors that appear in the console when starting the application?

I have eliminated all issues except one. I am not sure what to do with this last one: /public/assets/tests/empty-vo-widget.js

>   1:10  error  'empty' is defined but never used  no-unused-vars
>   1:19  error  Unnecessary semicolon              no-extra-semi

erikap:
Didn't look at the code yet, but clicking through the screens I encountered the following problems:
- contact info screen contains fixed value 'Hasselt' instead of the bestuurseenheid of the current session
- selecting birth date on the create person screen only allows a limited range of dates

mehran-naghizadeh:
> Didn't look at the code yet, but clicking through the screens I encountered the following problems:
> 
> * contact info screen contains fixed value 'Hasselt' instead of the bestuurseenheid of the current session
> * selecting birth date on the create person screen only allows a limited range of dates

I have fixed the first issue now. For the second one: since I have only used an existing component, I am not sure why it is so. However, I guessed this might be an intended feature to only allow people older than 18 years.

PR #13, opened 2020-02-06
Feature/accessibility quickfixes

Dietr:
Improves the accessibility of the interface:
- fixes the heading hiërarchy
- added a label to the account switch
- added a focus state to the modal close button


mietcls:
Great additions. One minor issue: there already is a class to visually hide elements from Webuniversum: u-visually-hidden — we can use that instead of creating u-hidden-visually. After that we can merge.

PR #28, opened 2020-12-07
Add data-test* attribute to submit form

aatauil:
`data-test-field-uri="submit-form-button"`


cecemel:
Cool! would you also be able to add a data-test* attribute 'maak nieuwe melding'?
thx

cecemel:
Don't we need the package in the dependencies, package.json

PR #35, opened 2021-02-17
If the address-suggestion is empty the adressenregister won't make a …

HugaertsDries:
…call to the back-end.


claire-lovisa (approved):
Reviewed and tested, working well ! :)

PR #36, opened 2021-02-26
Replace decprected route mixins (ember-simple-auth) 

aatauil:


nvdk:
:eyes:  following because we will want to do the same in GN

aatauil:
@nvdk 

The PR has a pretty good explanation on what to do to upgrade
https://github.com/simplabs/ember-simple-auth/pull/2198

The only thing that wasn't very clear was replacing the `ApplicationRouteMixin`. If you want custom logic for e.g. `sessionInvalidated`, you will now have to create a new service that extends from ember-simple-auth's SessionService and there you can overwrite that method with you custom logic :smile:

app/routes/subsidy/applications/index.js
```diff
@@ -1,8 +1,10 @@
 import Route from '@ember/routing/route';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import { inject as service } from '@ember/service';
 import DataTableRouteMixin from 'ember-data-table/mixins/route';
 
-export default class SubsidyApplicationsIndexRoute extends Route.extend(AuthenticatedRouteMixin, DataTableRouteMixin) {
+export default class SubsidyApplicationsIndexRoute extends Route.extend(DataTableRouteMixin) {
+  @service session;
```
Windvis, line 6: This doesn't seem to be used and the parent handles the check already.

app/routes/index.js
```diff
@@ -1,6 +1,13 @@
 import Route from '@ember/routing/route';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import { inject as service } from '@ember/service';
+
+export default class IndexRoute extends Route{
+  @service session;
+  @service currentSession;
+
+  async beforeModel(transition) {
+    this.currentSession.load()
```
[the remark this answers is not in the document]
Windvis: Also, I don't think this line needs to be executed if the session is not authenticated, right? So I would swap this below the simple-auth check, and wrap that one in an if?

app/services/session.js
```diff
@@ -0,0 +1,23 @@
...
+    const logoutUrl = ENV['torii']['providers']['acmidm-oauth2']['logoutUrl'];
+    if (logoutUrl.startsWith('http')) {
+      window.location.replace(logoutUrl);
+    }
+    else {
+      warn('Incorrect logout URL configured', { id: 'session-invalidation-failure' });
+    }
+  },
+
+  _loadCurrentSession() {
```
Windvis, line 16: I think this was not meant to be here?

app/services/session.js
```diff
@@ -0,0 +1,23 @@
+import SessionService from 'ember-simple-auth/services/session';
+import ENV from 'frontend-loket/config/environment';
+
+export default SessionService.extend({
+
+  handleInvalidation() {
+    const logoutUrl = ENV['torii']['providers']['acmidm-oauth2']['logoutUrl'];
+    if (logoutUrl.startsWith('http')) {
+      window.location.replace(logoutUrl);
```
Windvis, line 9: I think `super.handleInvalidation(logoutUrl)` should work as well and is a nicer solution imo :smile:

app/services/session.js
```diff
@@ -0,0 +1,23 @@
+import SessionService from 'ember-simple-auth/services/session';
+import ENV from 'frontend-loket/config/environment';
+
+export default SessionService.extend({
```
Windvis, line 4: I think a native class should work as well?

app/routes/index.js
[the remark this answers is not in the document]
Windvis: Actually, I think this line can go since the application route calls it already.

app/routes/application.js
```diff
@@ -1,39 +1,22 @@
...
   beforeModel() {
     const moment = this.moment;
     moment.setLocale('nl-be');
     moment.setTimeZone('Europe/Brussels');
     moment.set('defaultFormat', 'DD MMM YYYY, HH:mm');
-
     return this._loadCurrentSession();
   }
 
-  sessionAuthenticated() {
```
Windvis, line 22: I think this part is missing after the refactor, or did I overlook it?

Windvis (changes requested):

Windvis:
Optional, and it's probably a nice premature DRY refactor :smile: :

Since the session service is already extended, you could override the `requireAuthentication` and `prohibitAuthentication` methods as well and provide default route names.

app/controllers/subsidy/applications/index.js
```diff
@@ -1,7 +1,9 @@
 import Controller from '@ember/controller';
 import { SENT_STATUS } from '../../../models/submission-document-status';
+import { inject as service } from '@ember/service';
 
 export default class SubsidyApplicationsIndexController extends Controller {
+  @service() session;
```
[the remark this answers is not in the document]
aatauil: Yeah seems like it , the 'Application' probably threw me off guard there

app/routes/application.js
aatauil: https://github.com/simplabs/ember-simple-auth/pull/2202#issuecomment-676367340

Windvis: Yea sure, but the service has a method you can override, right? Or do you mean that it's no longer needed?

Windvis: (Btw, I have no background context on this project, so it might not actually be needed. I just wanted to make sure that it wasn't an oversight :smile:)

app/routes/index.js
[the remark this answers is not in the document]
aatauil: Yeah that semi-colon is actually important since emacs users are going to be mad at me xd
I think it suffices to place this.currentSession.load() below the requireAuthentication since if the session is not authenticated it will transition before exectuing that line ( I could be wrong though )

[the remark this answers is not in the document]
Windvis: I think it can be removed actually, since the application route already does it?

app/routes/switch-login.js
```diff
@@ -1,10 +1,13 @@
-import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 
-export default class SwitchLoginRoute extends Route.extend(UnauthenticatedRouteMixin) {
+export default class SwitchLoginRoute extends Route {
   @service() session;
 
+  beforeModel(){
+    this.session.prohibitAuthentication('login')
```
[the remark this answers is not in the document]
aatauil: Why would you think that maybe? SwitchLogin is it own route so if you would enter the url to that route in your browser then it should not let you visit that route if you are not logged in right?

app/services/session.js
aatauil: yeah but I just didnt want to mess with it right there and then and just focus on the mixins but indeed it has to be converted

app/services/session.js
aatauil: Will note it when changing to native class

app/services/session.js
aatauil: this was the most straightforward way of extending the sessionService without renaming it to a different class

app/routes/index.js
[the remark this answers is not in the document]
aatauil: No it actually gave me a lot of issues, for some reason unless actively calling it in the index route, when you would login using mock-login and it then transitions to index you will see that all modules in the index route will be 'disabled' until you reload the page

app/routes/subsidy/applications/index.js
aatauil: If not used it should be removed , I left it in since I though it was intentionally placed there for future use or smth? ( in context of authenticatedRouteMixin)

app/routes/switch-login.js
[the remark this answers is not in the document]
Windvis: Yea, but in that case it needs to be a `requireAuthentication` call instead, no? I assumed it was similar to the login page. If already logged in, redirect to index.

app/routes/supervision.js
```diff
@@ -1,10 +1,8 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
-
-export default class SupervisionRoute extends Route.extend(AuthenticatedRouteMixin) {
-
+export default class SupervisionRoute extends Route {
+  @service session;
```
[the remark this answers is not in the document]
aatauil: Yes indeed `    this.session.requireAuthentication(transition, 'login');` should be added to beforeModel()

app/routes/supervision/submissions/index.js
```diff
@@ -1,9 +1,16 @@
...
+import { inject as service } from '@ember/service';
+
+export default class SupervisionSubmissionsIndexRoute extends Route.extend(DataTableRouteMixin) {
+  @service() session;
 
-export default class SupervisionSubmissionsIndexRoute extends Route.extend(AuthenticatedRouteMixin, DataTableRouteMixin) {
   modelName = 'submission'
+
+  beforeModel(transition) {
+    this.session.requireAuthentication(transition, 'login');
```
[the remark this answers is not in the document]
aatauil: I am actually not 100% sure, so testing it without that line seeing if it acts as we want then it can be left out

app/routes/index.js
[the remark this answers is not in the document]
Windvis: I think that's the thing that the sessionAuthenticated in the Application route did before.

app/routes/application.js
aatauil: Actually this might be related to the issue I was having in the index route, so maybe override the method in the SessionService and remove the `this.currentSession.load()` on the index route and see if it will work? sorry I dont remember exactly what I already tried anymore

Windvis: Yea I think that should work!

app/routes/switch-login.js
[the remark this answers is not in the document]
aatauil: context: (You can only click the 'switchLogin` button from the index route , this to change to another "bestuurseenheid" since a user can belong to multiple bestuurseenheden) now for the moment I dont know exactly what it should do other then logging you in as part of a different bestuurseenheid, thats why I redirect to login. will need more info of Felix for this

aatauil:
@Windvis Yeah actually that would be nice, but at the same time it should always be flexible to accept other routes in case this is needed in the future ( who knows ). So probable a if (routeSpecified) else (goToDefaultRoute) if possible?

app/routes/switch-login.js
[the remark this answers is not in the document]
cecemel: Sorry ignorant person here, but the prohibit wasn't done in the previous version, why would we want this no?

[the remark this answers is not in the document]
Windvis: @cecemel The previous version included the `UnauthenticatedRouteMixin` mixin which is the equivalent. I don't think it would redirect to the login route though.

app/routes/supervision.js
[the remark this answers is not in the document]
Windvis: Done!

app/routes/switch-login.js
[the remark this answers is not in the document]
Windvis: @cecemel I don't know the context of the project so it's hard for me to judge if this change is valid.

The UnauthenticadedRouteMixin redirects a logged in user to the index page. So I _think_ this one needs to be changed to `this.session.prohibitAuthentication('index');`. But the current code will work since the login page also redirects to index when a user is logged in.

app/services/session.js
Windvis: Optional change

app/services/session.js
Windvis: Optional change

app/controllers/subsidy/applications/index.js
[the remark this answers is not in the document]
Windvis: Done!

app/routes/application.js
Windvis: @aatauil Did you give this a try?

aatauil: No it doesnt actually work, suddenly the index page acts up and everything just "breaks"... I thought it would work too tbh

Windvis: Ok! In that case: Done!

app/routes/index.js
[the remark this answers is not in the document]
Windvis: Done! (Related to discussion over here: https://github.com/lblod/frontend-loket/pull/36#discussion_r588268120)

app/routes/supervision/submissions/index.js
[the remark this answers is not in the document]
Windvis: Done!

app/services/session.js
Windvis: Done!

app/routes/subsidy/applications/index.js
cecemel: probably copy pasta, should not be there I think

app/routes/switch-login.js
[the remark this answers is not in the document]
cecemel: updated to index and tested

app/routes/application.js
cecemel: can someone ELI5 here :-) : by what is it replaced now?

cecemel:
The issue with switch-login, is known on chrome, where a  pop-up is blocked when coming  from a redirect. And will not be tackled here. (Unless someone has an idea) The plan would be to see how we can replace the popup by an iframe.

app/routes/application.js
Windvis: By the PR I made (and before that by calling `this.currentSession.load();`).

PR #37, opened 2021-03-12
Move the "current session" loading logic into the session service

Windvis:
I couldn't get it out of my head why it wouldn't work so I tried it as well :smile:.


Windvis:
@aatauil It does seem to work with the session service version :tada:

cecemel:
thanks (we have some old? problems with switching bestuurseenheid) but merging anway.

PR #38, opened 2021-03-17
Sort messages in berichtencentrum

claire-lovisa:


app/components/berichtencentrum/conversatie-view.js
```diff
@@ -1,12 +1,16 @@
...
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
+import { sort } from '@ember/object/computed';
 
 export default class BerichtencentrumConversatieViewComponent extends Component {
   @service() router;
 
+  @sort('berichten', 'generalSort') sortedBerichten;
+
+  @tracked generalSort = Object.freeze(['verzonden']);
```
HugaertsDries, line 11: What is this `Object.freeze` for?

HugaertsDries (approved):
Looks good, merge in to dev -> QA

app/components/berichtencentrum/conversatie-view.js
claire-lovisa: To ensure the sorting will only read the existing berichten and not modify then, not 100% it's required here now that you mention it

claire-lovisa: Ah from what I read [here](https://stackoverflow.com/questions/55431665/ember-avoid-leaking-state-in-ember-objects-error-for-sorting-property) now that we have native classes it's not needed anymore so i'll just remove it :)

PR #40, opened 2021-04-01
Rename all occurances of store.find

Windvis:
`store.find` is a private API method and only there to support Ember's default model data loading behavior.

It's just an alias for `store.findRecord` though, so an easy one to find and replace! :smile:


aatauil:
Looks good! everything still working as it should, thanks :)

PR #43, opened 2021-04-09
Replace soon to be deprecated transitionToRoute method

aatauil:
Replace all instances for this.transitionToRoute to this.router.transitionTo. Injecting the router service where needed ofcoarse


Windvis (approved):
Good stuff!

PR #46, opened 2021-05-03
Validate PR's with Drone

Windvis:
This setup validates the code and runs the tests, but they are all allowed to fail since the project isn't free from errors yet.
Once all linting errors are fixed the CI can be made more strict.


aatauil:
@cecemel Could you have a glance at this? I would actually really like to see this being implemented :D

PR #47, opened 2021-05-11
Embroider tracking

Windvis:
This PR includes the needed setup to run Embroider. There are a couple of blocking issues that need to be resolved before the `staticComponents` flag can be enabled. Once that flag is enabled, Embroider can be configured to split separate JS bundles for certain routes, which might be nice for the separate "modules" Loket has.

- ~~[ember-power-select](https://github.com/cibernox/ember-power-select/pull/1419) uses some dynamic component invocations~~ Should be fixed in the [v5 release](https://github.com/cibernox/ember-power-select/blob/master/CHANGELOG.md#504).
- ~~ember-power-select-create uses some dynamic component invocations. We'll inline the implementation since it causes other issues as well.~~ The addon received an update which resolved that issue.
- ~~ember-appuniversum has some issues with the `AuLink` component. Embroider doesn't like the use of `{{component "link-to"}}` but in theory that should work.~~ This is now implemented differently so it won't cause issues.
- ~~[@lblod/ember-submission-form-fields](https://github.com/lblod/ember-submission-form-fields/pull/20) also uses dynamic component invocations which needs to be wrapped in `ensure-safe-component`.~~ Resolved by https://github.com/lblod/ember-submission-form-fields/pull/53
- ~~[@lblod/ember-submission-form-fields](https://github.com/lblod/ember-submission-form-fields/pull/19) has some files that need to be deleted~~
- ~~the last release of torii includes a file that needs to be removed. [This is already solved](https://github.com/Vestorly/torii/pull/434), but not released. This is not really breaking since it doesn't fail the build when `staticAddonTrees` is enabled.~~ ember-acmidm-login v2 no longer uses torii.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -1,4 +1,4 @@
-import EmberRouter from '@ember/routing/router';
+import EmberRouter from '@embroider/router';
 import config from 'frontend-loket/config/environment';
 
 export default class Router extends EmberRouter {
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -47,6 +47,19 @@ module.exports = function(defaults) {
     staticAddonTrees: true,
     staticHelpers: true,
     staticComponents: true,
-    // splitAtRoutes: ['route.name'], // can also be a RegExp
+    splitAtRoutes: [
+      'mock-login',
+      'contact',
+      'legaal',
+      'mandatenbeheer',
+      'bbcdr',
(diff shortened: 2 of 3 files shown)


Windvis:
I will try to do some build size comparisons as well. Might be interesting to see if / how much the bundle size is decreased by using Embroider.

PR #48, opened 2021-05-17
Bump appuniversum to 0.4.3 (needed for the accordion component)

aatauil:
- Needed for the new accordion component and just needed a bump in general

### DIFF
--- a/app/templates/components/acmidm-login-compact.hbs
+++ b/app/templates/components/acmidm-login-compact.hbs
@@ -2,7 +2,7 @@
   {{yield}}
 </button>
 {{#if this.isAuthenticating}}
-  <AuLoader @size="small" />
+  <AuLoader @padding="small" />
 {{/if}}
 {{#if this.errorMessage}}
   <AuAlert @alertIcon="alert-triangle" @alertsize="small" @alertTitle={{this.errorMessage}} @alertskin="error" />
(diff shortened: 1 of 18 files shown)


Windvis:
There was a breaking change in the 0.4.0 release which needs to be handled I think:

 `<AuLoader @size>` was renamed to `@padding`.

https://appuniversum.github.io/ember-appuniversum/docs/atoms/au-loader

Windvis:
I think the new date picker also uses a different format by default so it probably requires updating the labels, or configure the datepicker.

aatauil:
@

> There was a breaking change in the 0.4.0 release which needs to be handled I think:
> 
> `<AuLoader @size>` was renamed to `@padding`.
> 
> https://appuniversum.github.io/ember-appuniversum/docs/atoms/au-loader

Not quite a "breaking" change as it is an easthetic thing but definitly something that should be dealt with. Actually missed that when going through the commits so good catch! Thanks you, should be ready to merge :)

Windvis:
I only know because I received the same feedback in GN :smile:. The appuniversum repo could really use a changelog to be honest.

Windvis (approved):

aatauil:
Yeah changelog or at least put a description on what changed when releasing (in the notes) since now it just says something like "release v1.0.0" . I will actually open an issue for it since upgrading would be much easier

PR #49, opened 2021-05-21
Update to ember 3.24 LTS

Windvis:
The 3.24 blueprint also includes a stricter "recommended" / "octane" ruleset, so this causes _a lot_ of extra linting errors.

eslint: 176 problems
template-lint: 311 problems

I don't think that's a bad thing since they all push the codebase towards modern Ember code, but it will take a while to fix them all.


aatauil (changes requested):
Changes look good , app builds as it should but running the app locally shows a deprecation warning in console:

```
deprecate.js:136 DEPRECATION: @linkRoute is deprecated, use @route instead [deprecation id: @appuniversum/ember-appuniversum.au-link.linkRoute-argument]
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:49519:21)
        at HANDLERS.<computed> (http://localhost:4200/assets/vendor.js:49652:9)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:49546:9)
        at HANDLERS.<computed> (http://localhost:4200/assets/vendor.js:49652:9)
        at invoke (http://localhost:4200/assets/vendor.js:49664:9)
        at Object.deprecate (http://localhost:4200/assets/vendor.js:49620:28)
        at AuLink.get route [as route] (http://localhost:4200/assets/vendor.js:141748:36)
        at getPossibleMandatoryProxyValue (http://localhost:4200/assets/vendor.js:27414:19)
        at _getProp (http://localhost:4200/assets/vendor.js:27479:17) 
```
Caused by the use of deprecated method in appuniversum. Since this upgrade is not urgent, I think it might be a better idea to address the deprecation in appuniversum first

Windvis:
@aatauil That deprecation is unrelated to the blueprint update. Appuniversum itself triggers it by design. AuLink now also supports `@route` so the `@linkRoute` arg is now deprecated. A simple find and replace should  fix it though.

PR #50, opened 2021-05-21
Refactor CurrentSession service

Windvis:
This refactors the currentSession service in a similar way as [in the GN app](https://github.com/lblod/frontend-gelinkt-notuleren/pull/81).

### DIFF
--- a/app/components/berichtencentrum/bericht-notificatie-voorkeuren.js
+++ b/app/components/berichtencentrum/bericht-notificatie-voorkeuren.js
@@ -4,24 +4,20 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 
 export default class BerichtencentrumBerichtNotificatieVoorkeurenComponent extends Component {
-  @service() currentSession;
+  @service currentSession;
 
-  @tracked wilMailOntvangen = this.currentSession.groupContent.wilMailOntvangen
-  @tracked emailAddress = this.currentSession.groupContent.mailAdres
+  @tracked wilMailOntvangen = this.currentSession.group.wilMailOntvangen
+  @tracked emailAddress = this.currentSession.group.mailAdres
 
   @action
     commit() {
-      // get all variables
-      let emailAddress = this.emailAddress;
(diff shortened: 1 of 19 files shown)


aatauil (commented):
Looked through every change and tested it thoroughly and eveything looks more then great! Since this change touches on auth I think its best to have another extra pair of eyes testing it again to be sure @cecemel @claire-lovisa @HugaertsDries

claire-lovisa:
When trying to open the Leidinggevendenbeheer module I get the following error 

```
Error while processing route: leidinggevendenbeheer.bestuursfuncties.index Assertion Failed: You attempted to access the `uri` property (of <(unknown):ember254>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('uri')` in this case.
```
It's, as the message suggests, going away when updating `bestuurseenheidClassificatie.uri` to `bestuurseenheidClassificatie.get(uri)`

Windvis:
@claire-lovisa, hmm I think that's because of this change: https://github.com/lblod/frontend-loket/pull/50/files#diff-49e04cc15d746fe1092a467dccffa7c1c84ef2964d22492fc8e3db8db1570400R9 . My bad!

That's an async relationship so I guess adding `await` there would work as well. An alternative could be to fetch and store the classification in the service as well. That has the benefit that you can access it in a sync way. The best solution is to make all relationships async and get rid of all the async relationship quirks, but that a looooot of refactoring :smile:.

I don't really have time in the foreseeable future to improve the PRs I made here though :disappointed:

claire-lovisa:
@Windvis thanks for the explanation :) So for now could we just add an `await` to be able to merge that PR, and we'll see the refactoring later on ?

Windvis:
I _think_ so. Storing it on the service is better though and a quick fix as well. I'll try to squeeze it in tomorrow :smile:

Windvis:
@claire-lovisa The problem should now be fixed, not sure how I missed that error last time :smile:.

PR #54, opened 2021-08-16
Chore/footer heading

aatauil:
- hostname now showing in footer heading.
- fix double scrollbar issue on help page.

Note: This will NOT work when using fastboot as it accesses the document object. For some reason I thought loket used fastboot but seems not to be the case?

### DIFF
--- a/app/components/shared/footer-heading.js
+++ b/app/components/shared/footer-heading.js
@@ -1,0 +1,7 @@
+import Component from '@glimmer/component';
+
+export default class SharedFooterHeadingComponent extends Component {
+  get hostname() {
+    return document.location.hostname;
+  }
+}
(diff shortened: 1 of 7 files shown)


app/components/shared/footer-heading.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@glimmer/component';
+
+export default class SharedFooterHeadingComponent extends Component {
+  get hostname() {
+    return document.location.hostname;
```
Windvis, line 5: In the other apps it's just hardcoded to the production url. Since it's not a link I don't think it really needs to be the same as the current host? (Even though that's more correct, of course).

Windvis: Feel free to merge if you prefer the dynamic version though.

aatauil: Yeah dynamic version has been approved by Felix, merging :)

PR #55, opened 2021-08-24
Chore/remove mirage

aatauil:
Delete ember-cli-mirage and mirage/ folder. There are acceptance tests written that use mirage but those are 2 years old and I doubt that they are still used. ember-test-selectors ( data-test.. in hbs ) I am pretty sure is being used by the testers so that should stay. If we decide not to use test then we can even consider remove test related packages and the test folder. (@ember/test-helpers, qunit, ...)


aatauil:
Apparently mirage is used by the droni-ci which I was unaware of. Closing this PR as mirage is being used

PR #56, opened 2021-08-30
Update subsidy forms initial status

claire-lovisa:
We use to put new forms in CONCEPT status. We now introduce a new status NEW for more ease when handling tailored data extraction.
A form becomes a CONCEPT when first saved.

### DIFF
--- a/app/controllers/subsidy/applications/edit/step/edit.js
+++ b/app/controllers/subsidy/applications/edit/step/edit.js
@@ -8,6 +8,7 @@ import { timeout } from 'ember-concurrency';
 import { dropTask, task } from 'ember-concurrency-decorators';
 import fetch from 'fetch';
 import { validateForm } from '@lblod/ember-submission-form-fields';
+import { CONCEPT_STATUS } from '../../../../../models/submission-document-status';
 
 export default class SubsidyApplicationsEditStepEditController extends Controller {
 
@@ -162,6 +163,7 @@ export default class SubsidyApplicationsEditStepEditController extends Controlle
       // NOTE update modified for the form and the consumption
       yield this.updateModified(this.semanticForm);
       yield this.updateModified(this.consumption);
+      yield this.updateStatus(this.semanticForm, CONCEPT_STATUS);
 
       this.updateRecentlySaved(); // TODO can this be done on a more "data" driven way
     } catch (exception) {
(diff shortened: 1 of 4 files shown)


app/controllers/subsidy/applications/edit/step/edit.js
```diff
@@ -162,6 +163,7 @@ export default class SubsidyApplicationsEditStepEditController extends Controlle
       // NOTE update modified for the form and the consumption
       yield this.updateModified(this.semanticForm);
       yield this.updateModified(this.consumption);
+      yield this.updateStatus(this.semanticForm, CONCEPT_STATUS);
```
cecemel, line 166: @claire-lovisa perhaps, for performance, would it be faster to do an if new -> then concept?

cecemel: This is for a later to think about, but i do find it strange now that i think of it, that the status is also managed by subsidy-application-management. I don't remember why.

cecemel (changes requested):

app/models/subsidy-application-flow-step.js
```diff
@@ -1,9 +1,9 @@
-import Model, { attr, belongsTo } from '@ember-data/model';
+import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 
 export default class SubsidyApplicationFlowStepModel extends Model {
   @attr order;
 
-  @belongsTo('file') formSpecification;
+  @hasMany('file') formSpecifications;
```
cecemel, line 6: This is rather surprising. In the model I though it would be only one.
https://github.com/lblod/app-digitaal-loket/blob/master/config/resources/master-subsidies-domain.lisp#L104

cecemel (changes requested):

app/models/subsidy-application-flow-step.js
cecemel: Which I understood, is the flowstep is the concrete instance of the subsidieprocedurestap. But sometimes forms change, without changing the abstract subisdieprocedurestap

cecemel: I reviewed the model, and I think the approach of the first bullet, was the intended path in changes. The fact that flows have been introduced, just complicates everything....

app/routes/subsidy/applications/edit/step/new.js
```diff
@@ -1,37 +1,45 @@
...
       creator: this.currentSession.user,
       lastModifier: this.currentSession.user,
       subsidyApplicationFlowStep: step,
       subsidyMeasureConsumption: consumption,
-      status: this.conceptStatus,
+      status: this.newStatus,
     });
 
-    form.sources.pushObject(spec);
+    form.sources.pushObject(latestSpec);
```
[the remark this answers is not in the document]
claire-lovisa: A set wont work as sources need to be an array I think

PR #59, opened 2021-10-15
Chore/refactor mandaten edit

aatauil:
Made a PR so we dont forget about it :p


Windvis:
I think it makes sense to close this PR. We probably will make changes to the module in the future to make it more consistent with the newer "eredienst mandatenbeheer" so this component will most likely be replaced anyways.

Windvis:
Closing since it's probably more work to fix the merge conflicts than it is to start over with these changes as a blueprint. The above mentioned reason will probably make this PR redundant eventually as well.

PR #62, opened 2021-11-17
Disable the subsidy deletion button when certain tasks are running

Windvis:
This adds a simple boolean flag to the consumption class which can be used to tag a consumption as stable or unstable. This flag is then used to disable the deletion button.

I also did some code restructuring to make it easier to read but it can be reverted if opinions differ :smile:.

### DIFF
--- a/app/controllers/subsidy/applications/edit.js
+++ b/app/controllers/subsidy/applications/edit.js
@@ -4,7 +4,6 @@ import { task } from 'ember-concurrency-decorators';
 import fetch from 'fetch';
 
 export default class SubsidyApplicationsEditController extends Controller {
-
   @service router;
 
   get reeksHasStartOrEnd(){
@@ -25,15 +24,19 @@ export default class SubsidyApplicationsEditController extends Controller {
   }
 
   @task
-  * delete() {
-    if (!this.canDelete) {
+  *delete() {
+    if (!this.canDelete || !this.consumption.isStable) {
       return;
     }
+
     try {
+      this.consumption.isStable = false;
       /**
        * NOTE: this endpoint prevents the removal of submitted forms, preventing the removal of a consumption all together.
        */
-      const forms = yield this.consumption.get('subsidyApplicationForms').toArray();
+      const forms = yield this.consumption
+        .get('subsidyApplicationForms')
+        .toArray();
       for (const form of forms) {
         yield fetch(`/management-application-forms/${form.id}`, {
           method: 'DELETE',
@@ -46,6 +49,8 @@ export default class SubsidyApplicationsEditController extends Controller {
       console.log('Removal of consumption failed because:');
       console.error(error);
       // TODO Error handling
+    } finally {
+      this.consumption.isStable = true;
     }
   }
 }
--- a/app/controllers/subsidy/applications/edit.js
+++ b/app/controllers/subsidy/applications/edit.js
@@ -19,8 +19,8 @@ export default class SubsidyApplicationsEditController extends Controller {
     return this.model.organization;
   }
 
-  get canDelete(){
-    return this.consumption.get('status.isConcept');
+  get canDelete() {
+    return this.model.consumptionStatus.isConcept;
   }
 
   @task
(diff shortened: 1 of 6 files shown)


app/routes/subsidy/applications/edit.js
```diff
@@ -20,11 +19,10 @@ export default class SubsidyApplicationsEditRoute extends Route {
       ].join(',')
     });
 
-    const organization = await this.session.groupContent;
```
Windvis, line 23: `groupContent` does no longer exist after the last CurrentSessionService changes so this wasn't working correctly.

app/controllers/subsidy/applications/edit/step/edit.js
```diff
@@ -174,13 +137,34 @@ export default class SubsidyApplicationsEditStepEditController extends Controlle
...
+    yield this.model.semanticForm.reload();
+    yield this.model.semanticForm.hasMany('sources').reload();
+  }
+
   @dropTask
   * submit() {
     try {
       if (this.canSubmit) {
+        this.preventConsumptionDeletion();
         yield this.save.perform();
```
Windvis, line 168: Hmm, it seems that the save task is also called in the submit task so this isn't probably not working 100% correctly yet. (The save task will set the `isStable` flag to true while the submit one is still running.)

This is the downside of the simple "flag" implementation, only 1 task that modifies the state can run at the same time..

app/controllers/subsidy/applications/edit.js
```diff
@@ -20,20 +19,24 @@ export default class SubsidyApplicationsEditController extends Controller {
     return this.model.organization;
   }
 
-  get canDelete(){
-    return this.consumption.get('status.isConcept');
+  get canDelete() {
+    return this.model.consumptionStatus.isConcept;
```
HugaertsDries, line 23: Why not make this:

```javascript
return this.model.consumptionStatus.isConcept && this.consumption.isStable

File: app/controllers/subsidy/applications/edit/step/edit.js [hunk 4]
```diff
@@ -111,77 +111,29 @@ export default class SubsidyApplicationsEditStepEditController extends Controlle
...
+      this.enableConsumptionDeletion();
+    }
   }
 
   @dropTask
   * submit() {
     try {
-      if (this.canSubmit) {
-        yield this.save.perform();
+      if (this.canSubmit && this.consumption.isStable) {
```
HugaertsDries, line 134: Like before, couldn't we move this to `this.canSubmit`?

app/models/subsidy-measure-consumption.js
```diff
@@ -1,6 +1,13 @@
 import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
+import { tracked } from '@glimmer/tracking';
 
 export default class SubsidyMeasureConsumptionModel extends Model {
+  // This flag can be used to mark this record "instable". It shouldn't be possible to delete instable records.
+  // Note that we assume there will only ever be 1 action that manipulates this value at the same time.
+  // If it needs to be possible to run actions concurrently this should replaced by an array
+  // of concurrency task instances and a getter which returns true if no tasks are running.
+  @tracked isStable = true;
```
HugaertsDries, line 9: This is honestly pretty cool. Clean and simple, the way @cecemel likes it.

HugaertsDries (approved):
Not much to add, pretty simple but effective solution. Approved.

app/controllers/subsidy/applications/edit.js
HugaertsDries: Because it is used in the template, don't want the button to disappear at every save. :heavy_check_mark:

Windvis: If `canDelete` is set to true the delete button will be hidden. In this case I assumed the button simply needed to be disabled when the tasks were running. Other than that no specific reason so it can be changed if that's the prefered UX :+1:.

app/controllers/subsidy/applications/edit/step/edit.js
Windvis: I did it like that first but I noticed [the generic error message](https://github.com/lblod/frontend-loket/blob/ba896d54a98086d51e0669f11775b5bd8dc906ec/app/templates/subsidy/applications/edit/step/edit.hbs#L105-L115) was then being shown while the tasks were running. It took me a while to figure out why that was cause I seemingly didn't touch any of that code :smile:.

So it can be done but more template refactoring would be needed and I wanted to keep this PR as simple as possible.

PR #64, opened 2022-01-14
Fix broken build because of a release of ember-get-config v1.0.0

claire-lovisa:
Context : https://github.com/embroider-build/embroider/issues/1077

Needed to use a new version of docker-ember embedding npm v8.x to use `overrides` in the package.json, as well as bumping some deprecated dependencies ;
```bumping ember-cli-sass to ^10.0.1 and running edi ember g ember-cli-sass (which will add the sass package to package.json) fixes the issue
Previous versions of ember-cli-sass use node-sass under the hood which is deprecated and doesn't work
with node 16.x. New versions of ember-cli-sass use sass (alias of dart-sass) and that works with node 16.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test"
   },
+  "overrides": {
+    "ember-get-config": "0.5.0"
+  },
   "devDependencies": {
     "@appuniversum/appuniversum": "0.0.15",
     "@appuniversum/ember-appuniversum": "0.4.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -25,6 +25,9 @@
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test"
   },
+  "overrides": {
+    "ember-get-config": "0.5.0"
```
Windvis, line 29: I would pin `@embroider/macros` to v0.50 instead. The real problem is that there are versions of the macros lower than v0.43.0 in the node_modules tree, which trigger the build error. Pinning it like this will most likely break again once more addons start updating the macros package.

PR #65, opened 2022-01-18
Fix the session setup code in acceptance tests

Windvis:
This mocks the "bestuurseenheid" that the test user belongs so the currentSession.load() method doesn't throw an exception which [immediately invalidates the session](https://github.com/lblod/frontend-loket/blob/5d7f1f1d5e27e58a788d0bb6f7cd439b296e3c1e/app/routes/application.js#L23) again. This fixes some of the acceptance tests.

### DIFF
--- a/tests/helpers/session.js
+++ b/tests/helpers/session.js
@@ -1,43 +1,62 @@
 import { authenticateSession } from 'ember-simple-auth/test-support';
 
-export const CLASSIFICATION_LABEL = "classificatieLabel";
+export const CLASSIFICATION_LABEL = 'classificatieLabel';
 
-const session = async function(server, options = {}) {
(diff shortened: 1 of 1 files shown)


tests/helpers/session.js
```diff
@@ -1,43 +1,62 @@
...
+
+  const classification = server.create('bestuurseenheid-classificatie-code', {
+    label: 'Test',
+  });
+
+  server.create('bestuurseenheid', {
+    naam: 'ABB',
+    id: '974816591f269bb7d74aa1720922651529f3d3b2a787f5c60b73e5a0384950a4',
+    classificatie: classification,
   });
```
Windvis, line 59: The actual new code, the rest is just formatting changes :smile:

PR #66, opened 2022-01-21
Add a "tiny" css modifier for the au-c-content class

Windvis:
Small PR which accompanies the backend PR: https://github.com/lblod/app-digitaal-loket/pull/122

This is used in the help text section of certain form fields. This class ensures the au-c-content looks similar to the help-text component.

image

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -500,3 +500,16 @@ $au-non-editable-background: $au-gray-100 !default;
     flex-grow: 1;
   }
 }
+
+// "tiny" au-c-content modifier which matches the size of the help-text component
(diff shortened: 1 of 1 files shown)


HugaertsDries (approved):
@Windvis Looks good. 
Only question I have: is there a way that we could recommend this to be added to app-universum? So that we could eventually move this out of the "shame" CSS.

Windvis:
Yes, I briefly discussed this with Dieter but he feels there might be better patterns than using so much content as help text. So the "good" solution will probably be to refactor to a different pattern but that required more work.

So I still went for this solution since the pattern was already in use and could use the quick improvement. It might mean that it won't be accepted in appuniversum though. We could still create an issue in the appuniversum repo to discuss that.

PR #67, opened 2022-02-07
Colocate all component template and js files

Windvis:
This change moves all component template files next to the js files using the codemod:

https://github.com/ember-codemods/ember-component-template-colocation-migrator


Windvis:
It seems the file upload components are overriding the addon version so they can't be moved to the components folder.

Windvis:
Putting this on hold until the appuniversum upgrade is done.

PR #68, opened 2022-02-08
Dependency maintenance

Windvis:
This updates some dependencies to the latest version and removes those that aren't used (anymore).

The mirage tests still fail, but the update was needed to unblock the ember-auto-import update. 

Since the tests aren't actually used, it might be better to simply remove mirage and use a different testing setup in the future.


package.json
```diff
@@ -26,87 +26,86 @@
...
+    "@glimmer/component": "^1.0.4",
+    "@glimmer/tracking": "^1.0.4",
     "@lblod/ember-acmidm-login": "1.3.0",
-    "@lblod/ember-mock-login": "^0.5.0-beta",
+    "@lblod/ember-mock-login": "^0.6.0",
     "@lblod/ember-submission-form-fields": "0.15.5",
     "@lblod/ember-toezicht-form-fields": "^0.6.1",
-    "@lblod/ember-vo-mu-file-upload": "^0.6.2",
-    "@lblod/ember-vo-webuniversum": "^0.22.7",
+    "@lblod/ember-vo-mu-file-upload": "^0.7.0",
```
Windvis, line 43: This breaks the file uploader because of the default endpoint change. https://github.com/lblod/ember-vo-mu-file-upload/commit/655d700e6bee14f4e3b8a0e5a46f378778a6057e

Windvis: We should roll this update back until we can update the form field library, but even then it might make more sense to simple switch to the appuniversum uploader instead.

PR #69, opened 2022-02-14
Help and contact page improvements

Windvis:


app/templates/help.hbs
```diff
@@ -1,158 +1,154 @@
-<main id="main">
-  <div class="au-c-main-container__content au-c-main-container__content--scroll" id="content">
-    <section class="au-o-region">
-      <div class="au-o-layout au-o-layout--small">
-        <AuHeading @level="1" @skin="1" class="au-u-margin-bottom">Helppagina</AuHeading>
-        <AuHeading @level="2" @skin="2" class="au-u-margin-bottom">Veelgestelde vragen</AuHeading>
+<div class="au-c-body-container au-c-body-container--scroll">
```
Windvis, line 1: This fixes the scroll issue when the user is logged in. This is a quick and dirty fix though. The application template renders a lot of extra html wrappers when authenticated which caused the current issue. It probably makes sense to revisit the setup and see how it can be improved.

PR #74, opened 2022-03-04
Use `AuFileUpload` instead of the webuniversum version

Windvis:
This removes the usage of ember-vo-mu-file-upload. It also requires a bump in the ember-submission-form-fields addons since it was also using the webuniversum file uploader.


package.json
```diff
@@ -37,9 +37,8 @@
     "@glimmer/tracking": "^1.0.4",
     "@lblod/ember-acmidm-login": "1.3.0",
     "@lblod/ember-mock-login": "^0.6.0",
-    "@lblod/ember-submission-form-fields": "0.15.5",
+    "@lblod/ember-submission-form-fields": "lblod/ember-submission-form-fields#refactor/modernize-components",
```
Windvis, line 40: This depends on a branch for now, until we get everything merged and released in that addon.

package.json
```diff
@@ -78,6 +77,7 @@
     "ember-modifier": "^3.1.0",
     "ember-moment": "^7.6.0",
     "ember-mu-transform-helpers": "^2.1.2",
+    "ember-named-blocks-polyfill": "^0.2.5",
```
Windvis, line 80: Once we update to 3.28 we can drop this dependency.

app/templates/components/berichtencentrum/conversatie-reactie.hbs
```diff
@@ -31,15 +31,26 @@
...
-      {{#each this.bijlagen as |bijlage|}}
-        <VoMuFileCard @file={{bijlage}} @onDelete={{this.deleteFile}} />
-      {{/each}}
-      <VoMuFileDropzone @onFinishUpload={{this.attachFile}} @maxFileSizeMB="1024" />
+    <div class="au-u-margin-bottom">
+      {{#if this.bijlagen}}
+        <ul class="au-o-flow au-o-flow--tiny au-u-margin-bottom-small">
+          {{#each this.bijlagen as |attachment|}}
+            <li>
+              <FileCard
```
Windvis, line 43: This component comes from the submission-form-fields addon at the moment since it was also needed there. Ideally this comes from appuniversum though: https://github.com/appuniversum/ember-appuniversum/issues/213

app/templates/components/bbcdr/report-edit.hbs
```diff
@@ -7,13 +7,29 @@
...
+
+      {{#if this.reportFiles}}
+        <ul class="au-o-flow--tiny au-o-flow au-u-margin-bottom-small">
+          {{#each this.reportFiles as |file|}}
+            <li>
+              <FileCard
+                @filename={{file.filename}}
+                @fileSize={{file.humanReadableSize}}
+                @downloadLink={{file.downloadLink}}
+                @onDelete={{unless @report.status.isVerstuurd (fn this.deleteFile file)}}
```
aatauil, line 19: I have a preference to put logic like this in the action itself instead of hidden in the hbs :/ completely subjective though

Windvis: I'm not sure I follow what you mean. This conditionally gives the component a handler function since the presence of that function is used to display the delete button. I don't really see a way to move this into the component class :smile:.

aatauil: You cant put it in a component class indeed, got confused :p
But wouldn't it be more 'clean' to have it as a separate argument. Now it basically depends on whether a callback is passed to onDelete. imo this is very logical but not as straight forward perhaps? I am doing some heavy mierenneuken here eh, just throwing an other perspective on the table
 
```
@showDelete={{eq @report.status.isVerstuurd false}}
@onDelete={{fn this.deleteFile file}}
```

But then again this could also be seen as overhead and bloat

Windvis: That's how it was in the previous version, but I found the extra argument redundant since it can be derived. So I axxed it :smile:. I find it confusing that @showDelete would be false while the `@onDelete` action is set :smile:. In my head the consumer wants to show a delete button if they provide a handler for it. But maybe this is just a matter of component design preference :smile:.

This is also how I implemented it in the appuniversum version: https://github.com/appuniversum/ember-appuniversum/pull/215 Maybe we should just merge this as-is for now and move the discussion there? This component is temporary anyway.

Windvis: I agree it's not very clear though (I hate `unless`). Maybe doing something like this would have been more readable? `@onDelete={{if @report.status.isVerstuurd null (fn this.deleteFile file)}}`

aatauil: Ohh I see yeah that does make sense but somehow I feel like there might be a very edgy  use case where you would want to show the delete without action. But it sound extremely unlikely unless it is some kind of hack :laughing: alright we will discuss there then, looking forward to the discussion :p

PR #75, opened 2022-03-07
Replace WuDatepicker with AuDatePicker

Windvis:
This builds on the #74 so that needs to be merged first.

This replaces the WuDatepicker component with the AuDatePicker equivalent. I added a temporary override for the dutch translations which can be removed once the following PR is merged in appuniversum: https://github.com/appuniversum/ember-appuniversum/pull/209


Windvis:
I'll move it into draft mode until the file upload PR is merged.

aatauil:
Some overflowing issues

**Mandatenbeheer**
bewerk > pas aan > wijzig ende mandaat
image

bewerk > pas aan > corrigeer fouten
image

Will try to find a appuniversum style fix for it and mark it in the code ;)

Windvis:
I think that's a known issue. I've seen @Dietr mention this before somewhere. The datepicker isn't rendered in a wormhole, so it has this problem in scrolling containers. Not sure if this is fixable from our side without changing datepicker implementations. The duet-datepicker repo is pretty silent as well..

The horizontal overflow is due to the data-table styles. (commenting `@import 'ember-vo-webuniversum-data-table';` should fix that, which will happen once we remove the webunversum addons anyway).

aatauil:
While looking for the fix I noticed that the class causing this sits in _shame.scss (disable the overflow rule fixes it) so lets just ignore and see if this fixes itself down the line
image

aatauil (approved):
LGTM! works as expected comparing side by side with dev environment. future note: overflow issue

Windvis:
Oh, I see, so the place where this is shown isn't scrollable, so you can't click all dates? Ok, yea that's a real problem, sorry, I misunderstood :smile:. Good catch, I didn't notice that when clicking through the app.

PR #83, opened 2022-03-15
Fix a problem where the subsidy deletion button was shown when it shouldn't be

Windvis:
This fixes the problem where the subsidy deletion button was shown after the subsidy was already submitted. The consumption status was never updated after the first load due to the way we stored the data. Making this state derived solves the issue.


Windvis:
It seems this just roll backs my [own changes](https://github.com/lblod/frontend-loket/pull/62/commits/02a01eef38c7e1d36bc3994f3c1fdc40d9622b12) :see_no_evil:. My bad. I really want us to get rid of all the async relationships though.

PR #86, opened 2022-03-21
Use node 16 and npm 8.3+ in CI

Windvis:
This matches our local setup and fixes the error in the lint:hbs step

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,26 +4,26 @@ type: docker
 name: test
 steps:
 - name: install
-  image: danlynn/ember-cli:3.20.0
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -4,26 +4,26 @@ type: docker
 name: test
 steps:
 - name: install
-  image: danlynn/ember-cli:3.20.0
+  image: danlynn/ember-cli:4.2.0-node_16.14
```
Windvis, line 7: This bundles ember-cli 4.2.0, but it will use our local version when running commands (and we run npm scripts anyways).

Issue #78, opened 2022-03-08
Ember 3.28 LTS update tracking


Windvis:
I gave upgrading to the 3.28 LTS a try since it should have been a quick bump and it would allow us to drop some polyfills, but I ran into some issues:

- ~~ember-moment needs an update since v7 isn't compatible with 3.27+ Upgrading to v9 works. v10 has also been released but that uses a different setup and since moment will be removed eventually v9 should be a good temporary version~~ fixed
- ~~_something_ caused a "too much recursion" error when accessing the leidinggevendenbeheer > functionaris > contact info page. I couldn't reproduce it anymore after a nuke of the node_modules folder but it's something we need to test properly.~~ can't reproduce it anymore
- ~~Saving a report on the BBC DR page triggers an Ember Data issue because we try to set a PromiseManyArray to a relationship while that needs to be resolved first. Resolving the relationship before using it should fix that issue. I wrote a more detailed explanation [here](https://github.com/lblod/frontend-gelinkt-notuleren/pull/200#issuecomment-1061962554), since the GN team ran into a similar issue during the upgrade. It's likely that more issues like this are happening so we need to test everything properly, or do a sweep for non-awaited relationships in the codebase.~~ pinned ember data on 3.27 for now

I made a [WIP branch](https://github.com/lblod/frontend-loket/pull/77/files) which solved some of the issues, but more work and testing is needed so I'm putting this on hold until the appuniversum update is done and there aren't more pressing things to do. I do think the update is important since it will help us when we start refactoring older components to the new system (named blocks, local helpers, ..).

Windvis:
Closing since the update has been merged, the Ember Data issues will be handled in a future PR.

PR #87, opened 2022-03-30
Update the BBC DR report file list when switching reports

Windvis:
This fixes the problem where switching between reports didn't update the list of uploaded files.

The updated code also fixes the issue we noticed [with Ember Data 3.28](https://github.com/lblod/frontend-loket/issues/78) so that's one less thing to worry about when doing the update.

### DIFF
--- a/app/components/bbcdr/report-table.hbs
+++ b/app/components/bbcdr/report-table.hbs
@@ -15,11 +15,11 @@
         {{! Actions }}
       </th>
     </c.header>
-    <c.body data-test-loket='bbcdr-report-table-body' as |row|>
+    <c.body data-test-loket='bbcdr-report-table-body' as |report|>
       <td data-test-loket='bbcdr-report-table-files-column'>
-        {{#if row.files}}
+        {{#if report.files}}
           <ul>
-            {{#each row.files as |file|}}
+            {{#each report.files as |file|}}
               <li data-test-loket="bbcdr-file-miniature">
                 {{!-- TODO: Add back a file miniature once appuniversum has a solution for this --}}
                 {{file.filename}} <span class="au-c-info-text">{{file.miniatureMetadata}}</span>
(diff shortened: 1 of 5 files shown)


app/components/bbcdr/report-table.hbs
```diff
@@ -31,14 +31,14 @@
...
+      <td class="au-u-visible-medium-up">{{report.lastModifier.fullName}}</td>
+      <td class="au-u-visible-medium-up">{{moment-format report.modified 'DD-MM-YYYY'}}</td>
       {{/unless}}
       <td class="au-u-visible-small-up">
-        {{shared/document-status-pill row.status}}
+        {{shared/document-status-pill report.status}}
       </td>
       <td data-test-loket='bbcdr-report-table-open-details-column'>
-        <LinkTo @route='bbcdr.rapporten.edit' @model={{row}} class="au-c-link">Bekijk</LinkTo>
+        <LinkTo @route='bbcdr.rapporten.edit' @model={{report.id}} class="au-c-link">Bekijk</LinkTo>
```
Windvis, line 41: Passing the model as an object prevents Ember from running the model hook, which we need to happen for the code to work properly.

app/routes/bbcdr/rapporten/edit.js
```diff
@@ -1,13 +1,14 @@
...
-        'status'
-      ].join(',')
+  async model(params) {
+    let report = await this.store.findRecord('bbcdr-report', params.id, {
+      include: ['files', 'last-modifier', 'status'].join(','),
     });
+
+    return {
+      report,
+      reportFiles: await report.files,
```
Windvis, line 11: This ensures that the relationship is resolved and we can use it safely in sync code. This is one of the things that caused issues in the Ember Data 3.28 update.

aatauil (approved):
Great stuff! This can be merged :)

PR #102, opened 2022-04-22
Remove the ember-cli-mirage testing setup

Windvis:
The setup was no longer working and getting it to work again is very time consuming for little benefit. The tests were alle failling and outdated as well.

Since the testing team is working towards full end-to-end testing it's probably better to simply drop the setup. We can always add a new modern setup in the future if needed or revert this PR and start from there.


mirage/config.js
```diff
@@ -1,37 +0,0 @@
...
-    Shorthand cheatsheet:
-
-    this.get('/accounts/:id/gebruiker');
-    this.get('/bestuurseenheden/:id');
-
-
-    http://www.ember-cli-mirage.com/docs/v0.4.x/shorthands/
-   */
-  this.get('/accounts/:id/gebruiker');
-  this.get('/bestuurseenheden/:id');
```
Windvis, line 24: This endpoint was giving issues. It seems the inflector config wasn't properly detected which made it look for `bestuurseenheden` instead of `bestuurseenheid`. Other models had the same issue so fixing that everywhere would not be simple. I couldn't find any issues in the repo, so I assume most people simply use english words which will not have the issue.

PR #104, opened 2022-04-25
Use browser-update directly

Windvis:
ember-browser-update isn't really needed anymore in modern Ember apps.


app/app.js
```diff
@@ -10,4 +11,11 @@ export default class App extends Application {
   Resolver = Resolver;
 }
 
+browserUpdate({
+  vs: { i: 11, f: -3, o: -3, s: -3, c: -3 },
```
Windvis, line 15: This seems to be the old API, but I can't really find docs about it so I kept it as-is. Ideally this is replaced with the `required` version.

PR #106, opened 2022-04-26
Resolve `AuAlert` deprecation warnings

Windvis:
The `alert*` prefixed argument names were deprecated in [v1.0.8](https://github.com/appuniversum/ember-appuniversum/blob/master/CHANGELOG.md#v108-2022-04-25).


Windvis:
I noticed there were a lot of cases were the incorrect casing was used, for example `@alertSize` vs `@alertsize`. This caused the component to fall back to the default value. After switching to the new argument name it will display the correct state again.

PR #110, opened 2022-05-04
Implement the "public services" overview page

Windvis:
This adds the public services overview page with some mock data.

The filters are mocked as well, pagination and sorting doesn't work since that was harder to mock and I didn't want to spend more time on it.

A lot of TODO comments will be resolvable once the mu-cl-resources config is added.

### DIFF
--- a/app/controllers/public-services/index.js
+++ b/app/controllers/public-services/index.js
@@ -1,0 +1,62 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { restartableTask, timeout } from 'ember-concurrency';
+
+export default class PublicServicesIndexController extends Controller {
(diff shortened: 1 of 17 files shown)


Windvis:
A lot of TODO comments will be resolvable once the mu-cl-resources config is added.

app/controllers/public-services/index.js
```diff
@@ -0,0 +1,62 @@
...
+    return this.isLoading;
+  }
+
+  get hasResults() {
+    return this.publicServices.length > 0;
+  }
+
+  get hasErrored() {
+    return this.model.loadPublicServices.isError;
+  }
```
Windvis, line 48: I think we can extract a more generic solution from this. I've used a similar setup in OP, so it might make sense to spend some time and make the whole "skeleton loading" + loading states use-case nicer to use.

PR #112, opened 2022-05-06
Implement the "add service" page

Windvis:
This allows users to add new services. At the moment all data is mocked.


app/mock-data/public-services.js
```diff
@@ -0,0 +1,78 @@
+export const sectors = [
```
Windvis, line 1: This file can  be removed once we switch to the proper backend implementation.

PR #115, opened 2022-05-11
Implement the "Product of dienst" details page

Windvis:
This adds the main structure for the page and displays the mock data where needed.

### DIFF
--- a/app/controllers/public-services/details.js
+++ b/app/controllers/public-services/details.js
@@ -1,0 +1,17 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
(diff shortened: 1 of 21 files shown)


app/controllers/public-services/details.js
```diff
@@ -0,0 +1,17 @@
...
+export default class PublicServicesDetailsController extends Controller {
+  @service router;
+
+  @action
+  removePublicService() {
+    // TODO: replace this with a backend call
+    let index = addedMockPublicServices.indexOf(this.model.publicService);
+    addedMockPublicServices.splice(index, 1);
+
+    this.router.replaceWith('public-services');
```
Windvis, line 15: I'm having some issues with the `router.replaceWith` method that needs further research but I've spent too much time on it already. At the moment the history entry isn't actually replaced..

PR #116, opened 2022-05-16
Feature/accountability table

aatauil:
New subsidy Table for ukraine subsidy form. 
To be tested in sync with https://github.com/lblod/app-digitaal-loket/pull/164

Important:
- for the address, select an address with busnummer for the hidden busnummer dropdown to show
- make sure table rows are capped at 20 rows 
- for details see jira & miro


app/components/rdf-form-fields/accountability-table/edit.hbs
```diff
@@ -0,0 +1,49 @@
+<AuLabel>
+  Vul per opvangplaats de adresgegevens en het aantal slaapkamers in. Per opvangplaats moet je de facturen of bewijsstukken voor opknapwerken toevoegen
+</AuLabel>
+<AuLabel for={{this.inputFor}}>
```
aatauil, line 4: You might notice that `@error={{this.hasErrors}}` is missing (show the red circle when table has errors). there is a very strange error when adding this. Tried to debug but very difficult so left it out at the moment since its not critical

Windvis: Do we even need this label? I don't see the matching form element and inputFor doesn't seem to return anything?

app/components/rdf-form-fields/accountability-table/edit.js
```diff
@@ -0,0 +1,320 @@
...
+    return this.errors.length == 0;
+  }
+
+  get isInvalid() {
+    return !this.isValid;
+  }
+}
+
+export default class RdfFormFieldsAccountabilityTableEditComponent extends InputFieldComponent {
+  id = `accountability-table-${guidFor(this)}`;
```
Windvis, line 65: Is this used somewhere?

app/components/rdf-form-fields/accountability-table/table-row.hbs
```diff
@@ -0,0 +1,114 @@
...
+          @onFinishUpload={{this.addFile}}
+          @disabled={{@disabled}}
+        />
+        <br>
+    {{/unless}}
+     {{#each this.files as |fileField index|}}
+      {{#if fileField.record}}
+        {{#let fileField.record as |file|}}
+          <li>
+            <FileCard
```
Windvis, line 95: The `AuFileCard` component should be available in our current appuniversum version I think, so I would use that.

Windvis: The Miro mentions that users can potentially upload 10 files, and I don't see that working with the current implementation 🤔. I wonder if a "card" pattern wouldn't have been better instead of a table.

Windvis (changes requested):

Windvis:
I also noticed that the address select shows "undefined" sometimes.

I can consistently reproduce it by adding a row, selecting an address and saving it. Everything still works as expected. Add a new row and temp save again. Reloading the page will show "undefined" in the row that was added first.

image

app/components/rdf-form-fields/accountability-table/edit.hbs
aatauil: good point, I confused it with a fieldGroup header for some reason. This has been removed :)

Windvis:
Prettier isn't happy either 😄 running `npm run lint:js:fix` should fix that, but I suggest you try to integrate eslint in your editor somehow so it can scream at you before committing 😬.

app/components/rdf-form-fields/accountability-table/edit.hbs
```diff
@@ -0,0 +1,49 @@
+<AuLabel>
```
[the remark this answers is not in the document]
aatauil: It can be yes, AuLabel is used just for the styling of it. (code comes from application for table)  Other tables use the same so keeping it for the sake of consistenty

app/components/rdf-form-fields/accountability-table/edit.js
aatauil: similar to {{yield}}, residue from application form code

app/components/rdf-form-fields/accountability-table/edit.js
```diff
@@ -0,0 +1,313 @@
...
+      });
+      const file = files.get('firstObject');
+      if (file) return new FileField({ record: file, errors: [] });
+      else
+        return new FileField({
+          record: null,
+          errors: ['Geen bestand gevonden'],
+        });
+    } catch (error) {
+      warn(`Failed to retrieve file with URI ${uri}: ${JSON.stringify(error)}`);
```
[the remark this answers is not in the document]
aatauil: This code comes from the files component in ember-submission-form-fields
https://github.com/lblod/ember-submission-form-fields/blob/071ec363b201b4af1dc527d65e94fdd2ab6ad82b/addon/components/rdf-input-fields/files.js#L94-L114

Could indeed be removed but I dont think it is urgent at this time

app/components/rdf-form-fields/accountability-table/table-row.hbs
```diff
@@ -0,0 +1,114 @@
+<tr>
+  <td class="au-u-1-3 {{if this.addressErrors "ember-power-select--error"}}">
```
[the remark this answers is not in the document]
aatauil: Oups, very nice catch. This will be removed. Since there are 2 power-selects in this <td/>, each had to be wrapped. Forgot to remove the old one, thanks :)

app/components/rdf-form-fields/accountability-table/edit.hbs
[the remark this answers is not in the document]
Windvis: I _really_ don't understand that mentality. Just because someone did something wrong before doesn't mean we need to make the same mistake in new code 😄.

There are some typografy utils which can be used to match the label styling if it _needs_ to look like that. I'm not sure if it actually does though, and basic text styles would probably be good enough.

app/components/rdf-form-fields/accountability-table/edit.hbs
```diff
@@ -0,0 +1,49 @@
...
+  Vul per opvangplaats de adresgegevens en het aantal slaapkamers in. Per opvangplaats moet je de facturen of bewijsstukken voor opknapwerken toevoegen
+</AuLabel>
+<AuLabel for={{this.inputFor}}>
+  {{@field.label}}
+</AuLabel>
+{{#each this.errors as |error|}}
+    <AuHelpText @error={{true}}>{{error.message}}</AuHelpText>
+{{/each}}
+<table class="data-table data-table--zebra data-table--tight au-u-margin-top-small">
```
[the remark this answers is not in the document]
aatauil: Hmm can we poor this into a story and immediately tackle all tables in custom form components at the same time? Doing it here and leaving the others the same would seem weird to the user I feel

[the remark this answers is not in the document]
Windvis: There already is [an issue](https://binnenland.atlassian.net/browse/DL-3484) for that, but it is not prioritized so it probably will never happen unless I go an a maintenance spree again. Which is why I would prefer not adding more tables in that style but I understand that the  feature is time sensitive, so your / Felix's call.

PR #119, opened 2022-05-31
Add a semantic form setup

Windvis:
This adds a form rendering setup to the tab pages. This requires the [feature/lpdc-management-service](https://github.com/lblod/app-digitaal-loket/compare/feature/lpdc-management-service) branch in the backend repo to work.

Things that will / might be done in separate PRs:
- show a confirmation modal before leaving if there are unsaved changes
- testing that persisting changes works
- disable the delete button while another action is running: https://github.com/lblod/frontend-loket/pull/119/files#diff-913ddcd8c6ea307624761ad98b570d2f6f797ac5fb2d17bbcaac632b2bb419eaR76


aatauil (approved):
Works :) made me realize that we need bodyparser in lpdc-management service since we use application/json instead of application/vnd.api+json when saving forms. Thanks for that :p

aatauil:
We should keep note of the temp github ref as source in package.json and not forget to fix that later

Windvis:
Thanks for the review!

> We should keep note of the temp github ref as source in package.json and not forget to fix that later

I don't think we'll forget since we need to merge the addon PR and release a new version anyways.

PR #120, opened 2022-06-03
Prepare release 0.61.0

cecemel:
It seems a lot of incremental changes has been done (+extra components). I bundle these here so these can be tested to QA


cecemel:
FYI @aatauil @Windvis @FangioD

PR #122, opened 2022-06-10
Allow users to create new public service instances

Windvis:


aatauil (approved):
LGTM, small question: What do you think of using a loading substate like for subsidies? I am guessing we would add this at a later stage?

app/templates/public-services/add.hbs
```diff
@@ -1,115 +1,123 @@
...
-        </td>
-      </tr>
-    {{/if}}
-  </Table.content>
-</AuDataTable>
\ No newline at end of file
+      {{/if}}
+    </Table.content>
+  </AuDataTable>
+{{else}}
```
Windvis, line 118: @aatauil I experimented with the separate "new" page, similar to the subsidy setup, but it added unneeded complexity imo. Passing the correct id to the route and making sure it wasn't accessible as a history entry aren't things I want to deal with, especially since I noticed some issues with the router service's "replaceWith" method.

I think keeping it in this route is better and simpler (at the cost of a slightly larger template).

aatauil: Alright, main reason for me asking was that that is the way its done in subsidies and I really like consistency. Would still love to address it later on but for now, we can roll with it

Windvis: Well, I would be more inclined to change the subsidy system over to this version.

Windvis: Ah yea, sorry, I didn't mean _now_ 😅 . But if we want to go for consistency I would prefer to change the subsidy version since I think the route based setup is strange 😄. But we can discuss pros / cons when we get there. 👍

PR #123, opened 2022-06-10
Delete public service

Windvis:
Builds on #122 so that needs to be reviewed and merged first.


app/components/public-services/confirm-deletion-modal.hbs
```diff
@@ -0,0 +1,37 @@
+<AuModal
```
Windvis, line 1: This wasn't designed but I thought it was risky to allow users to delete a public service with a single click.

app/components/public-services/confirm-deletion-modal.hbs
```diff
@@ -0,0 +1,37 @@
...
+  @modalOpen={{true}}
+  @closeModal={{this.close}}
+  data-test-confirm-deletion-modal="modal"
+>
+  <:title>Product of dienst verwijderen</:title>
+  <:body>
+    <div class="au-c-content">
+      <p>
+        TODO: Add a proper warning text
```
Windvis, line 10: This needs copy, but I'm not the best writer 😄. Do we have similar examples in other projects already?

app/components/public-services/confirm-deletion-modal.hbs
aatauil: I agree! nice addition :)

app/components/public-services/confirm-deletion-modal.hbs
aatauil: Yeah unless there is some extra info the user needs to know about like (deleted service can be recovered in 30 days, after that they will be permanently be deleted) something like that, but that is not the case here. For the title a simple 'Bent u zeker dat u dit product permanent wilt verwijderen? or something in that line will suffice

aatauil:
Weird thing happening in chromium browsers. When clicking "Annuleer" in the "verwijderen" modal, after it closes you cannot interact with the app anymore. Did some digging and found this: https://github.com/simplabs/ember-promise-modals/issues/665

Windvis:
> Weird thing happening in chromium browsers. When clicking "Annuleer" in the "verwijderen" modal, after it closes you cannot interact with the app anymore. Did some digging and found this: https://github.com/simplabs/ember-promise-modals/issues/665

It sounds like the "inert" feature that was rolled out in newer browsers: https://github.com/appuniversum/ember-appuniversum/issues/273

The fix for that was released in ember-appuniversum 1.1.1 so you probably need to purge your node_modules and npm install again to get the latest version.

aatauil:
> The fix for that was released in ember-appuniversum 1.1.1 so you probably need to purge your node_modules and npm install again to get the latest version.

That fixed it :D

aatauil:
I am actually not sure. For create and delete we are using ember-data. So for delete I kinda expected ember-data to remove the record from the list of products when it gets deleted (without backend call). Currently in the backend when you delete a publicService it will first return a statusCode and only then delete the service. I can make it wait but might be slow

PR #124, opened 2022-06-15
public services details page improvements

Windvis:
Includes #123 so that needs to be merged first.

This:
- moves the form save button back to the toolbar since the current version is confusing
- shows a confirmation modal if there are unsaved changes when going to a different page / tab.


app/components/public-services/details-page.hbs
```diff
@@ -0,0 +1,62 @@
...
+      </AuButton>
+
+      <AuButton
+        @skin="secondary"
+        @disabled={{or
+          (not this.hasUnsavedChanges)
+          this.saveSemanticForm.isRunning
+        }}
+        @loading={{this.saveSemanticForm.isRunning}}
+        form="test"
```
Windvis, line 38: woops, this still needs to be updated

app/components/public-services/details/unsaved-changes-modal.hbs
```diff
@@ -0,0 +1,41 @@
+<AuModal @modalOpen={{true}} @closeModal={{this.cancel}}>
+  <:title>Wijzigingen bewaren?</:title>
+  <:body>
+    <p class="au-c-content">
+      Er zijn velden gewijzigd. Wil je deze bewaren voor je naar een andere
+      sectie gaat? Indien niet dan worden deze wijzigingen niet bewaard.
```
Windvis, line 6: I just copied this from the design, let me know if it needs to be adjusted.

aatauil (approved):
LGTM, found a very small issue (used model instead of args when deleting) but fixed it as its a quick fix.

PR #126, opened 2022-06-17
Add basic publish handling

Windvis:
This adds a basic publish handling setup.

The missing things are:
- ~~the actual API call to publish the public service~~
- some sort of [notifications system](https://tailwindui.com/components/application-ui/overlays/notifications). The subsidy module uses simple alerts, but I think a "notification" system would be better since we can also show messages after transitioning. The message is usually also shown in the top right corner so we don't have the issue that it is not visible due to the scroll position. Kaleidos has something like that, but ideally we add support for it to Appuniversum. I created https://github.com/appuniversum/ember-appuniversum/issues/297 to track that.

### DIFF
--- a/app/components/public-services/details-page.hbs
+++ b/app/components/public-services/details-page.hbs
@@ -7,13 +7,14 @@
       <AuLoader @padding="small" />
     </div>
   {{else}}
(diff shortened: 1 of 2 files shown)


Windvis:
I rebased and included the changes from #146 with some extra tweaks.

PR #128, opened 2022-06-30
Feature/lpdc module

cecemel:
opening the PR


Windvis:
Hmmm, it seems the build is failing because [ember-appuniversum 1.3](https://github.com/appuniversum/ember-appuniversum/releases/tag/v1.3.0) dropped the dedicated appuniversum package but the rdfa-editor adds config that depends on that: https://github.com/lblod/ember-rdfa-editor/blob/29c948eff52a1d44c1391083ecd8f79d0eecba19/index.js#L36-L39

PR #132, opened 2022-07-12
Replace browser-rdflib with rdflib v2

Windvis:
rdflib v2 is compatible with browsers so we no longer need browser-rdflib (which is based on rdflib v1).

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@glimmer/tracking": "^1.0.4",
     "@lblod/ember-acmidm-login": "^1.4.0",
     "@lblod/ember-mock-login": "^0.7.0",
-    "@lblod/ember-submission-form-fields": "^1.8.0",
+    "@lblod/ember-submission-form-fields": "^1.9.0",
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
     "browser-update": "^3.3.38",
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "@glimmer/tracking": "^1.0.4",
     "@lblod/ember-acmidm-login": "^1.4.0",
     "@lblod/ember-mock-login": "^0.7.0",
-    "@lblod/ember-submission-form-fields": "^1.8.0",
+    "@lblod/ember-submission-form-fields": "git@github.com:lblod/ember-submission-form-fields.git#2b55969ce4f74c4db67578ac8bf74d8143fc9328",
```
Windvis, line 41: This depends on https://github.com/lblod/ember-submission-form-fields/pull/84 which in turn depends on https://github.com/redpencilio/forking-store/pull/11 and https://github.com/lblod/submission-form-helpers/pull/17. So we need a couple of releases before merging this 😄.

Windvis:
Most things seem to work fine. 1 major blocker is that rdflib throws an error when trying to save a subsidy form.

https://github.com/linkeddata/rdflib.js/blob/c14dfd57d5159ad5ac1ee2523cc7924968e24f53/src/serializer.js#L30 This line of code fails when Ember's array prototype extensions are enabled. The array will have a `_super` key which throws an exception and stops the form submission.

This means that we would first need to disable the prototype extensions in the app, which can be a bit of work. All places where we use things like `pushObject` or `removeObject` need to explicitly be wrapped in an Ember array.

PR #134, opened 2022-07-15
Set the base app title

Windvis:
This ensures that the base application title is set so it's displayed when navigating into a module. This provides more context for screenreaders and bookmarks.

Before: "Personeelsbeheer"
After: "Personeelsbeheer | Loket voor lokale besturen"


Windvis:
Small PR, but it bothered me 😬.

PR #138, opened 2022-07-20
Changed mandatatis rangorde to simple string

benjay10:
This PR makes the attribute type of the rangorde of a mandataris a simple string instead of a `language-string` from `ember-mu-transform-helpers`.

This is due to a bug in Virtuoso https://github.com/openlink/virtuoso-opensource/issues/1055, where some triples with language tags are not removed when you are updating the mandataris rangorde. The first value mu-cl-resources can find is displayed, which is not necessarily the latest value. With these changes, the value is a simple string and these are properly removed when updating to a new value.

This PR requires https://github.com/lblod/app-digitaal-loket/pull/185.

### DIFF
--- a/app/components/mandatenbeheer/mandataris-edit.js
+++ b/app/components/mandatenbeheer/mandataris-edit.js
@@ -77,10 +77,7 @@ export default Component.extend({
       this.get('mandataris.beleidsdomein').setObjects(this.beleidsdomeinen);
       this.set('mandataris.start', this.startDate);
       this.set('mandataris.einde', this.endDate);
-
-      if (this.rangorde) this.set('mandataris.rangorde', this.rangorde);
-      else this.set('mandataris.rangorde', undefined);
-
+      this.set('mandataris.rangorde', this.rangorde);
       this.set('mandataris.status', this.status);
 
       return yield this.mandataris.save();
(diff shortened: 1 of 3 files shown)


app/models/mandataris.js
```diff
@@ -1,7 +1,7 @@
 import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 
 export default class MandatarisModel extends Model {
-  @attr('language-string') rangorde;
+  @attr('string') rangorde;
```
Windvis, line 4: The `string` transform isn't needed if the backend already sends it like that.

```suggestion
  @attr rangorde;
```

app/components/mandatenbeheer/mandataris-edit.js
```diff
@@ -78,11 +78,7 @@ export default Component.extend({
       this.set('mandataris.start', this.startDate);
       this.set('mandataris.einde', this.endDate);
 
-      if (this.rangorde)
-        this.set('mandataris.rangorde', {
-          content: this.rangorde,
-          language: 'nl',
-        });
+      if (this.rangorde) this.set('mandataris.rangorde', this.rangorde);
       else this.set('mandataris.rangorde', undefined);
```
Windvis, line 82: I don't think the `if / else` is needed anymore and it can simply become `this.set('mandataris.rangorde', this.rangorde);` like the other mandataris properties?

PR #140, opened 2022-07-25
PersoonSearchFormPagination linting and bugfix

Poltergeistz:
When triggering a search with Zoek Voornaam and browsing above page 1, if an other search is triggered with input Familienaam, `this.currentPage` shouldn't keep the previous value.

This is causing errors and missleads the user (e.g in image 2 : The vorige button shouldn't be displayed as the page is 0 but since `this.currentPage` is keeping values from the first search this is the expected result). 

Page 2 - Loket voor lokale besturen

Page 1 - Loket voor lokale besturen

- Resolves the todo no-action.

- PersoonSearchFormPagination linter fix.

- `this.currentPage` fix

### DIFF
--- a/app/components/shared/persoon/persoon-search-form-pagination.js
+++ b/app/components/shared/persoon/persoon-search-form-pagination.js
@@ -2,18 +2,8 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 
 export default class PersoonSearchFormPagination extends Component {
-  classNames = ['data-table-pagination'];
-
-  constructor() {
-    super(...arguments);
-  }
-
   get currentPage() {
-    if (this.args.page) {
-      if (this.args.links.self.number !== this.args.page) return 1;
-      return parseInt(this.args.page) + 1;
-    }
-    return 1;
+    return this.args.page ? parseInt(this.args.page) + 1 : 1;
   }
 
   get firstPage() {
--- a/app/components/shared/persoon/persoon-search-form-pagination.js
+++ b/app/components/shared/persoon/persoon-search-form-pagination.js
@@ -1,50 +1,60 @@
-/* eslint-disable ember/no-classic-components, ember/no-classic-classes, ember/require-tagless-components, ember/require-computed-property-dependencies, ember/no-get, ember/require-computed-macros, ember/no-actions-hash */
-import Component from '@ember/component';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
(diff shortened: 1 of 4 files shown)


app/components/shared/persoon/persoon-search-form-pagination.js
```diff
@@ -1,50 +1,60 @@
...
-  firstPage: computed('links', function () {
-    return this.get('links.first.number') || 1;
-  }),
-  lastPage: computed('links', function () {
-    const max = this.get('links.last.number') || -1;
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class PersoonSearchFormPagination extends Component {
+  classNames = ['data-table-pagination'];
```
Windvis, line 5: classNames isn't used anymore in Glimmer components since it doesn't add an implicit wrapping div.

This worked like this in classic components:

```hbs
<div class="data-table-pagination">
{{! rest of the template that's visible in the persoon-form-pagination.hbs file}}
</div>
```

So to make the Octane version match the Ember component version that wrapping div should be added to the template file. If it doesn't affect the styling it might not be needed, but I would still delete the `classNames` property since it doesn't do anything anymore.

app/components/shared/persoon/persoon-search-form-pagination.js
```diff
@@ -1,50 +1,60 @@
...
+
+  constructor() {
+    super(...arguments);
+  }
```
Windvis, line 9: Can be removed since it's not used.

app/components/shared/persoon/persoon-search-form-pagination.js
```diff
@@ -1,50 +1,60 @@
...
+
+  get currentPage() {
+    if (this.args.page) {
+      if (this.args.links.self.number !== this.args.page) return 1;
```
Windvis, line 13: As mentioned in the chat, I think the real fix is to let the PersoonSearchForm component reset the page to 0 when it triggers a new search. That way this component will receive the new value as an argument and the template should rerender.

Windvis (changes requested):
Left some small remarks.

Thanks for working on the Octane refactoring! :tada:

app/components/shared/persoon/persoon-search-form-pagination.js
Poltergeistz: Made the change, looks like it makes no difference with or without the class. I'm going to remove it.

Windvis (approved):

PR #142, opened 2022-07-25
Convert dates in timezones with date-fns-tz

benjay10:
When selecting the start or end date of a mandate, the datepicker will always choose midnight in the timezone setting of the user. This is not correct, because we want to specify a date in the timezone of the administrative unit, namely Belgium. Using date-fns-tz we can convert and offset the date.

This was tested by setting the system timezone to a different timezone (e.g. East Australia as UTC+10) and inspecting the date in the patch requests. Setting the start/end date of a mandate should *always* produce a date such as `2022-05-31T22:00:00Z` (or 23:00 for dates outside of summer time) to indicate `2022-06-01`.

I chose to use `date-fns` because this is used in Appuniversum and `moment` is being phased out in a lot of projects.


cecemel:
@benjay10 (@Windvis ) 
Thanks
I am not sure this is the most user friendly way of doing this. I think now the user has to calculate the difference manually. furthermore when rendering  back; it still will have weird results for computers in other timezones

app/components/mandatenbeheer/mandataris-edit.js
```diff
@@ -246,7 +247,9 @@ export default Component.extend({
     },
 
     handleDateChange(propName, isoDate, date) {
-      this.set(propName, date);
+      //Convert to desired (Belgian) timezone
+      const converted = zonedTimeToUtc(date, 'Europe/Brussels');
```
Windvis, line 251: Another possible solution is to create a custom Ember Data transform which does this logic before serializing the date. That would make it easier to reuse since I assume this problem will occur in other places where we use the date picker? I assume it can simply extend the `datetime` transform from ember-mu-transform-helpers and override the serialize method?

benjay10: I was thinking the same thing, but it felt odd to create a "date in Belgian timezone" as a reusable transform. Especially because I think a good transform helper works in 2 ways and parsing the date from Ember Data was already done with Moment here.

What's really at fault here in my opinion is that the datepicker assumes the user's timezone, which is usually the correct thing. Ideally, I would like to set a timezone override on that component.

cecemel:
@Windvis @benjay10 can we pick this up again. If possible, could we do it case by case? So e.g. a date transform, but only for mandatarissen and leidinggevenden?

benjay10:
Won't be changed; discussed with @cecemel and @Windvis.

PR #144, opened 2022-07-27
Add a feature flag for the "eredienst mandatenbeheer" module

Windvis:
This adds a basic feature flagging system.

It should allow us to merge the `feature/eredienstMandatarissen` branch without having to enable the feature on dev.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -39,6 +39,7 @@ module.exports = function (environment) {
         },
       },
     },
+    features: {},
   };
 
   if (environment === 'development') {
--- a/config/environment.js
+++ b/config/environment.js
@@ -39,7 +39,10 @@ module.exports = function (environment) {
         },
       },
     },
-    features: {},
+    features: {
+      'eredienst-mandatenbeheer':
+        '{{FEATURE_EREDIENSTEN_MANDATENBEHEER_ENABLED}}',
+    },
   };
 
   if (environment === 'development') {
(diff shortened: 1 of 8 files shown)


config/environment.js
```diff
@@ -39,6 +39,10 @@ module.exports = function (environment) {
         },
       },
     },
+    features: {
+      'eredienst-mandatenbeheer':
+        '{{FEATURE_EREDIENSTEN_MANDATENBEHEER_ENABLED}}',
```
Windvis, line 44: We can replace this with `"true"` on the server when we want the feature enabled.

tests/unit/helpers/is-feature-enabled-test.js
```diff
@@ -0,0 +1,37 @@
...
+
+    config.features.foo = 'true';
+    assert.true(
+      isFeatureEnabled('foo'),
+      "it returns true if the value is the string 'true'"
+    );
+
+    assert.false(
+      isFeatureEnabled('baz'),
+      "it returns false if the feature name doesn't exist"
```
Windvis, line 32: Should we throw an error during development instead?

config/environment.js
cecemel: Would it be an option to also toggle for build versions? Like -  https://github.com/mu-semtech/ember-proxy-service#configure-the-frontends-variables

Windvis: That's what this does already? But maybe I misunderstood.

At the moment the feature will be "enabled" if the flag has the value `"true"` or `true`. 

For this PR that would mean the feature is enabled if we set `EMBER_FEATURE_EREDIENSTEN_MANDATENBEHEER_ENABLED` to `"true"` on the server. By default it will be disabled.

tests/unit/helpers/is-feature-enabled-test.js
Windvis: I went with the assert version since that will help us to find typos more easily.

config/environment.js
cecemel: check

Windvis:
Part of this setup has already been merged into master by #154

PR #146, opened 2022-07-29
Adding publish function

Poltergeistz:
Adding publish function for the API call to the publish service.

The modal button Verzend naar Vlaamse overheid now triggers the `publishPublicService` function.


Windvis:
I included your changes in #126 so this can be closed.

PR #147, opened 2022-07-29
Added projectname to index table of module subsidy

FangioD:
Together with https://github.com/lblod/app-digitaal-loket/pull/192

Working code: project name is added for fietsinfrastructuur subsidy.
However code should be reviewed:

- Currently located in model, however not that clean, better suggestions?
- Suggestions on sorting possibilities
- bit dirty code

### DIFF
--- a/app/models/subsidy-application-form.js
+++ b/app/models/subsidy-application-form.js
@@ -17,8 +17,7 @@ export default class SubsidyApplicationForm extends Model {
   })
   modified;
 
-@attr projectName;
-  
+  @attr projectName;
 
   // TODO faze out
   // @belongsTo('contact-punt') contactinfo; // default needed
--- a/app/models/subsidy-application-form.js
+++ b/app/models/subsidy-application-form.js
@@ -17,6 +17,9 @@ export default class SubsidyApplicationForm extends Model {
   })
   modified;
 
+@attr projectName;
+  
+
   // TODO faze out
   // @belongsTo('contact-punt') contactinfo; // default needed
   // @belongsTo('bank-account') bankAccount; // default needed
--- a/app/models/subsidy-measure-consumption.js
+++ b/app/models/subsidy-measure-consumption.js
@@ -28,4 +28,13 @@ export default class SubsidyMeasureConsumptionModel extends Model {
       );
     return undefined;
   }
+
+  get projectName(){
+    let projectNames = new Set();
+    for ( let form of this.subsidyApplicationForms.toArray() ){
+      if (form.projectName)
+        projectNames.add(form.projectName);
+      }
+    return [...projectNames].join(', ');
+  }
 }
--- a/app/models/subsidy-measure-consumption.js
+++ b/app/models/subsidy-measure-consumption.js
@@ -28,12 +28,4 @@ export default class SubsidyMeasureConsumptionModel extends Model {
       );
     return undefined;
   }
-
-  get projectName() {
-    let projectNames = new Set();
-    for (let form of this.subsidyApplicationForms.toArray()) {
-      if (form.projectName) projectNames.add(form.projectName);
-    }
-    return [...projectNames].join(', ');
-  }
 }
--- a/app/models/subsidy-measure-consumption.js
+++ b/app/models/subsidy-measure-consumption.js
@@ -29,12 +29,11 @@ export default class SubsidyMeasureConsumptionModel extends Model {
     return undefined;
   }
 
-  get projectName(){
+  get projectName() {
     let projectNames = new Set();
-    for ( let form of this.subsidyApplicationForms.toArray() ){
-      if (form.projectName)
-        projectNames.add(form.projectName);
-      }
+    for (let form of this.subsidyApplicationForms.toArray()) {
+      if (form.projectName) projectNames.add(form.projectName);
+    }
     return [...projectNames].join(', ');
   }
 }
--- a/app/templates/subsidy/applications/index.hbs
+++ b/app/templates/subsidy/applications/index.hbs
@@ -45,12 +45,14 @@
 
     <c.body as |consumption|>
       {{#let consumption.subsidyApplicationFlow.subsidyMeasureOfferSeries as |series|}}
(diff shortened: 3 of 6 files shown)


app/models/subsidy-application-form.js
```diff
@@ -17,6 +17,8 @@ export default class SubsidyApplicationForm extends Model {
   })
   modified;
 
+  @attr projectName;
```
Windvis, line 20: I noticed some strange behaviour due to this change. Changing the projectName in step 1 doesn't seem to work consistently.

After saving the form we [reload the Ember Data record](https://github.com/lblod/frontend-loket/blob/80a7803bfba069dc362ba0c398b41797bbb6c570/app/controllers/subsidy/applications/edit/step/edit.js#L190-L193) to fetch the new data that was entered in the form. This however seems to depend on some delay to work, because if it happens to fast the previous projectName is returned.

We then [update the `modified` attribute](https://github.com/lblod/frontend-loket/blob/80a7803bfba069dc362ba0c398b41797bbb6c570/app/controllers/subsidy/applications/edit/step/edit.js#L197) and save the record again. Since the projectName is now part of the record, the old value is sent to the backend and the new project name gets overwritten again.

A possible solution is to create a custom serializer class for the `subsidy-application-form` model in which we [disable the projectName serialization](https://api.emberjs.com/ember-data/4.6/classes/JSONAPISerializer/properties/attrs?anchor=attrs) since we only want to update this from the rdf forms.

Windvis: Another thing I noticed is that if you do succeed in changing the project name, it won't be propagated to the other steps until the user loads that step's form.

For example.

- change the projectName in step 1 from `project X` to `project Y`.
- go back to the overview
- Overview will show `project X, project Y` as project name
- click bewerk
- go to step 2
- go to overview
- Overview will show `project Y` as project name

So it seems something is syncing the data between the steps, but it only happens when the form is requested?

app/models/subsidy-measure-consumption.js
```diff
@@ -28,4 +28,12 @@ export default class SubsidyMeasureConsumptionModel extends Model {
       );
     return undefined;
   }
+
+  get projectName() {
+    let projectNames = new Set();
```
Windvis, line 33: I'm not 100% sure we need to show multiple names? It seemed to me that the projectName is supposed to be the same between all the steps? I don't have much background context about this module so I might be misunderstanding things.

cecemel: We are working in two data-model-spaces. 

At one side; we have the one respecting the data.vlaanderen, which you could consider as a kind of data model skeleton covering the general subsidy model. -> meant for data re-use.

On the other side; we have the 'dirty' space; every type of subsidy has multiple steps; which each contain a form.
That form basically covers the information about the subject of the subsidy (e.g. noodopvang, fietspaden etc)
  - A wide area of topics is covered here, they are not properly modeled given that
    -   The information gathered here is just the minimal information someone on the back office needs to make the decision yes/no you can get your grant. No further data re-use OF THE FORMS (yet)
    -  It would take ages if you want to properly model every subject of a subsidy.

The data of the forms is not interconnected here; so basically every form has its own little data world. 

The field projectName is one of them, but as the uses cases become clear; contact-person should also be one of them. And bankrekening too. 

So far we accepted these lived as dirty data; since the users in the backoffice got their excel of report (and don't really care about data) and we had to move on.

But now the problem becomes a bit more obvious; since we want to re-use this data in the fronted and we bump into funky stuff. 

So time is ripe to start thinking on how to tackle this problem. 
 - We need an overview of standard fields useful for ABB; which should be attached subsidiemaatregelconsumptie:
 -   - persoondata, bankrekeningnummer, projectnaam (perhaps more 'uw referentie')
 - think how we can capture this information (not necessarily though semantic forms)

But that's longer term; we don't have a lot of room now. But it is certainly something the business will appreciate (at least the people managing loket); because we can make very tailored UIX patterns for this general information.
If you want to have a call about it let me know.

Anyway; for now KISS it;
- don't show project naam on level of the subsidiemaatregel consumptie; but on application-form
-   pick the first one you encounter in the template
eg. `consumptie.applicationForms[0].projectName`

It will help the end users; if they complain it is not consistent; we have extra leverage to make this better.

cecemel: @Windvis @FangioD 
ok my KISS approach is really too flaky. I would suggest to ignore that for now.

app/templates/subsidy/applications/index.hbs
```diff
@@ -45,7 +45,13 @@
 
     <c.body as |consumption|>
       {{#let consumption.subsidyApplicationFlow.subsidyMeasureOfferSeries as |series|}}
-        <td>{{consumption.subsidyMeasureOffer.title}}</td>
+        <td>{{consumption.subsidyMeasureOffer.title}}
+           <AuHelpText @skin="secondary" class="au-u-margin-top-none">
+            {{#if (eq consumption.subsidyMeasureOffer.title "Fietsinfrastructuur")}}
```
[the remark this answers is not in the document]
cecemel: @FangioD this should be fixed.

app/models/subsidy-application-form.js
cecemel: @FangioD this should be fixed

app/models/subsidy-measure-consumption.js
cecemel: We need more time to clean up the data model. See business communications.

app/models/subsidy-application-form.js
Windvis: I disabled the projectName serialization so the first remark is fixed.

Windvis:
I rebased on development so a reset to the new remote commits is needed if you are still on a local branch based on the old history.

app/models/subsidy-measure-consumption.js
Windvis: We now show the projectName of the first form in the relationship which might be the wrong one, but the best we can do until we restructure the data.

PR #148, opened 2022-08-01
Fix all template lint errors

Poltergeistz:
Don't know if html comments we're used as a reminders.

The 37 todos left are from #59 and we can call it done for now.


Windvis (approved):
Thanks! 👍

PR #149, opened 2022-08-01
Ember-cli-babel override

Poltergeistz:
This is a work around for torii deprecation warnings.


package.json
```diff
@@ -26,6 +26,7 @@
     "test:ember": "ember test"
   },
   "overrides": {
+    "ember-cli-babel": "^7.26.11",
```
Windvis, line 29: It's also possible to make it match the version we have defined in the devDeps list with the following config:
```suggestion
    "ember-cli-babel": "$ember-cli-babel",
```

This has the benefit that we don't have to bump it when we bump the devDeps version. I _assume_ it works for devDeps as well, the docs only mention dependencies.

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

PR #151, opened 2022-08-04
Feature/eredienst bedienaren

cecemel:


Windvis:
I merged development so the feature flag system is now available 👍.

app/models/minister-position-function.js
```diff
@@ -0,0 +1,5 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class MinisterPositionFunctionModel extends Model {
+  @attr('string') label;
```
Windvis, line 4: Transforms should only be used if the backend sends the wrong type. In this case a string is already sent so we can drop the transform.
```suggestion
  @attr label;
```

Windvis (commented):
First iteration of feedback. I'll also jump in and help change the remarks I made.

PR #153, opened 2022-08-05
added a fix for validations running on page load

Asergey91:
this fixes an issue where fields with the rdfa editor were validated on page load (unlike other fields)

### DIFF
--- a/app/components/rdf-form-fields/rich-text-editor.js
+++ b/app/components/rdf-form-fields/rich-text-editor.js
@@ -39,6 +39,7 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
     //rdfaEditor setup is async, updateValue may be called before it is set
     if (this.editor) {
       const editorValue = this.editor.htmlContent;
+      // hack to prevent validations from firing on page load
       if (!this.firstRun) {
         super.updateValue(editorValue);
       }
--- a/app/components/rdf-form-fields/rich-text-editor.js
+++ b/app/components/rdf-form-fields/rich-text-editor.js
@@ -6,13 +6,6 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
   editor;
   inputId = 'richtext-' + guidFor(this);
 
-  constructor() {
-    super(...arguments);
-
-    // The editor returns an empty string if it contains no content, so we default to that as well.
-    this.value = '';
-  }
-
   get editorOptions() {
     return {
       showToggleRdfaAnnotations: false,
@@ -56,6 +49,15 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
       if (this.value !== editorValue) {
         super.updateValue(editorValue);
       }
+    }
+  }
+
+  loadProvidedValue() {
+    super.loadProvidedValue();
+
+    if (this.value == null) {
+      // The editor returns an empty string if it contains no content, so we default to that to make the value comparison check works as expected.
+      this.value = '';
     }
   }
 }
--- a/app/components/rdf-form-fields/rich-text-editor.js
+++ b/app/components/rdf-form-fields/rich-text-editor.js
@@ -5,7 +5,14 @@ import SimpleInputFieldComponent from '@lblod/ember-submission-form-fields/compo
 export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFieldComponent {
   editor;
   inputId = 'richtext-' + guidFor(this);
-  firstRun = true;
+
+  constructor() {
+    super(...arguments);
+
+    // The editor returns an empty string if it contains no content, so we default to that as well.
+    this.value = '';
+  }
+
   get editorOptions() {
     return {
       showToggleRdfaAnnotations: false,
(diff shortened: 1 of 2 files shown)


cecemel:
@Windvis to clarify: It seems the editor throws focusOut events after init. This is an issue in the editor that needs investigating. 
This codes should prevent the validation occurs immediately after load

cecemel:
@Asergey91 perhaps in code, you can explain this too

Windvis:
@cecemel Thanks for the info, I was just going to ask about that 😄.

Although, instead of fixing the `focusOut` event being triggered I think the real fix is that the editor provides a proper "The content was updated" event which we can use to trigger the update, which according to [this comment](https://github.com/lblod/ember-submission-form-fields/pull/78#discussion_r884822520) seems to be on the roadmap already?

cecemel:
@Windvis yeah indeed that would be a better abstraction. Could you perhaps open the discussion on the chat (#core-editor)?

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -39,6 +39,7 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
     //rdfaEditor setup is async, updateValue may be called before it is set
     if (this.editor) {
       const editorValue = this.editor.htmlContent;
+      // hack to prevent validations from firing on page load
```
cecemel, line 42: Can you just elaborate a bit: because issue in editor, will need to fixed on that level. (So we know in the future a bit more about the why? rather than the what/how?)
In my opinion comment should first tackle the why? and if needed (because complex algorithm on the what/how)

cecemel (changes requested):

cecemel:
@Windvis if @Asergey91 fixes my last comment, I am ok with merge. You?

Windvis:
Yes, sure! We don't really have any other options at the moment I think 😄.

Windvis:
We discussed this issue a bit in the Rocket.Chat and while there now is a event we can use to detect content changes it seems it still has some issues which prevents us from using it here. We'll merge this as a temporary fix for now.

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -39,7 +39,13 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
     //rdfaEditor setup is async, updateValue may be called before it is set
     if (this.editor) {
       const editorValue = this.editor.htmlContent;
-      super.updateValue(editorValue);
+      // hack to prevent validations from firing on page load
+      // this is due to a behaviour in the editor that fires a focusout event on init
+      // TODO: remove this once its fixed in the editor
+      if (!this.firstRun) {
+        super.updateValue(editorValue);
```
Windvis, line 46: I wonder, would it also make sense to compare the new value with the old one before triggering the update? If we combine that with the `@stealFocus={{false}}` argument I think it covers most false positives?

I did some debugging to find out why some editors were still being focused and it seems to be caused by the browser APIs that are used to write the data to the DOM. So not something that is (easily) fixable I guess.

Comparing the values should prevent the case where a user clicks the editor to trigger an update as well.

Windvis: I went with this approach. With the previous version it was possible to miss content changes due to the `firstRun` check. If for some reason the `focusOut` event wasn't triggered on init for an editor instance, and the user added actual content, the `firstRun` was still set to true so the update would be skipped. By comparing the current value to the old one we sidestep that issue.

Windvis:
I force pushed to this branch since I rebased on dev. If you have the previous version checked out you will need to reset it to the origin version.

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -29,7 +36,11 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
   @action
   handleRdfaEditorInit(editor) {
     this.editor = editor;
-    editor.setHtmlContent(this.value ? this.value : '');
+
+    // We only set the value if it contains actual content. This prevents the browser from focusing the editor field due to the content update.
+    if (this.value) {
+      editor.setHtmlContent(this.value);
+    }
```
Windvis, line 43: Not strictly needed since we do the value comparison anyways, but it seemed like unnecessary work for the editor.

Asergey91:
uhm it doesnt seem to load the data into the editor on my machine anymore.
on page reload all editor instances are blank even though im sure those fields are filled in.

Windvis:
Yikes, it seemed to work when I tried it 😄. I'll check after the demo.

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -6,6 +6,13 @@ export default class RdfFormFieldsRichTextEditorComponent extends SimpleInputFie
   editor;
   inputId = 'richtext-' + guidFor(this);
 
+  constructor() {
+    super(...arguments);
+
+    // The editor returns an empty string if it contains no content, so we default to that as well.
+    this.value = '';
```
Windvis, line 13: It's this 😄, must have made a last minute change without testing. Sorry, my bad.

Windvis: Should be fixed now.

Asergey91:
looks good

PR #160, opened 2022-08-12
added ui for backend validations

Asergey91:


cecemel:
@Asergey91 looks functional thanks!
@Windvis can you glance at the code?

app/components/public-services/details-page.hbs
```diff
@@ -88,4 +88,15 @@
       <AuButton @skin="secondary" {{on "click" this.sending}}>Annuleer</AuButton>
     </AuButtonGroup>
   </:footer>
-</AuModal>
\ No newline at end of file
+</AuModal>
+<AuModal @modalOpen={{this.showSubmitErrorModal}} @size="small">
```
Windvis, line 92: Hmm, it would have been nice if this used the `modal` service instead since I generally think it it's a better solution compared having to manage the modal state in the component itself. 

I don't think it's a blocker for this PR though since it works and it seems the "sending" popup also uses this setup already. Just mentioning in case you add more modal features in the future.

app/components/public-services/details-page.js
```diff
@@ -194,11 +212,17 @@ async function saveFormData(serviceId, formId, formData) {
...
-  await fetch(`/lpdc-management/${serviceId}/submit`, {
+  const response = await fetch(`/lpdc-management/${serviceId}/submit`, {
     method: 'POST',
     body: JSON.stringify({}),
     headers: {
       'Content-Type': 'application/json; charset=UTF-8',
     },
   });
+  if (response.status == 500) {
+    throw `Unexpected error during validation  of service "${serviceId}".`;
```
Windvis, line 223: We don't seem to catch this so this will make the app behave unexpectedly. Should we show a different message in the modal if this happens?

app/components/public-services/details-page.js
```diff
@@ -92,24 +105,29 @@ export default class PublicServicesDetailsPageComponent extends Component {
...
+
+      if (errors.length == 0) {
+        this.router.transitionTo('public-services');
+      } else if (errors.length == 1) {
+        //TODO: should probably handle this more in a more user friendly way
+        //ie: redirect to said form and scroll down to the first invalid field
+        this.showSubmitErrorModal = true;
+        const formId = errors[0].form.id;
+        this.submitErrorMessage =
+          'Er zijn fouten opgetreden in de tab "' + FORM_MAPPING[formId] + '". Gelieve deze te verbeteren!';
```
Windvis, line 124: nitpick but template literals generally are considered easier to read
```suggestion
          `Er zijn fouten opgetreden in de tab "${FORM_MAPPING[formId]}". Gelieve deze te verbeteren!`;
```

app/components/public-services/details-page.js
```diff
@@ -13,6 +13,12 @@ import ConfirmDeletionModal from 'frontend-loket/components/public-services/conf
 import UnsavedChangesModal from 'frontend-loket/components/public-services/details/unsaved-changes-modal';
 import { loadPublicServiceDetails } from 'frontend-loket/utils/public-services';
 
+//TODO: this is a bad idea this is the third time (as far as i know) these ids have been hardcoded
```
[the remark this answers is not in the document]
Asergey91: i mean ideally its source should be the form ttls them selves.
however they have ext:form as a uri atm
and the actual uris/ids are hard coded in 2 places in the frontend and 1 place in the service
but this should be a separate story i think

app/components/public-services/details-page.js
Asergey91: I mean if this gets ever hit we can notify the user, but at that point there is nothing the user can do... but i guess it couldnt hurt

app/components/public-services/details-page.hbs
Asergey91: i have no idea that even existed, ill check it out

app/components/public-services/submit-error-modal.hbs
```diff
@@ -0,0 +1,15 @@
+
+<AuModal 
+@modalOpen={{this.showSubmitErrorModal}}
```
Windvis, line 3: I would hardcode it to `true`

app/components/public-services/submit-error-modal.js
```diff
@@ -0,0 +1,13 @@
...
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class PublicServicesSubmitErrorModalComponent extends Component {
+  @tracked submitErrorMessage;
+  @tracked showSubmitErrorModal = true;
+
+  @action
+  closeSubmitErrorModal() {
+    this.args.close();
```
Windvis, line 11: I would use this as `@close` in the template so the backing class isn't needed.

app/components/public-services/details-page.hbs
```diff
@@ -89,14 +89,3 @@
     </AuButtonGroup>
   </:footer>
 </AuModal>
-<AuModal @modalOpen={{this.showSubmitErrorModal}} @size="small">
```
Windvis: 🎉 Thanks, although you didn't have to do it for this PR 😄.

Windvis (approved):

PR #161, opened 2022-08-18
Model updates; attributes removal; features removed

cecemel:
some updates on the model lead to removal of feaures

Do we get rid of https://github.com/lblod/frontend-loket/blob/development/app/components/public-services/sector-select.hbs or do we keep it as an example, because probably we will need similar things on other selectors


app/routes/public-services/index.js
```diff
@@ -23,23 +19,16 @@ export default class PublicServicesIndexRoute extends Route {
...
-      sector,
     };
   }
 
   @restartableTask
-  *loadPublicServicesTask({ search, sector, page, sort }) {
+  *loadPublicServicesTask({ search, page, sort }) {
     let query = {
       'page[number]': page,
-      include: 'sectors,status,lifecycle-status',
```
Windvis, line 42: I think we still want to include the `status` relationship since it's shown in the table?

cecemel: He sees it all :-) fixing it

app/components/public-services/status.js
```diff
@@ -1,34 +1,22 @@
...
-    'Ontwerp',
-  'http://data.lblod.info/id/concept/ipdc-thema/27d784f8-e7ea-4f18-88f8-02d427033009':
-    'Actief',
-  'http://data.lblod.info/id/concept/ipdc-thema/12873bfb-d2cc-4cc5-9255-1bf0ca2747f9':
-    'Stopgezet',
-};
+const DOCUMENT_STATUS = {
+  'http://lblod.data.gift/concepts/79a52da4-f491-4e2f-9374-89a13cde8ecd': {
+    skin: 'warning',
+    label: 'ontwerp',
```
Windvis, line 6: It's intended that these labels no longer start with an uppercase letter?

Windvis (approved):

Windvis:
There is still a linting error but I think it's resolved by the other PR so might not be worth fixing for this one?

app/templates/public-services/add.hbs
```diff
@@ -90,8 +75,6 @@
       {{else if this.hasResults}}
         <Content.body as |publicService|>
           <td>{{! template-lint-disable no-triple-curlies~}} {{{publicService.nameNl}}}</td>
-          <td>{{publicService.identifier}}</td>
```
Windvis, line 93: oh, I think we can also remove the attr from the model?

cecemel:
> There is still a linting error but I think it's resolved by the other PR so might not be worth fixing for this one?

yeah about a hasMany import? Should be fixed in the following pr

PR #163, opened 2022-08-22
change status on save

Asergey91:


Asergey91:
@felix the status concepts that were being used for publicservice.status are the same ones as publicservice.lifecycleStatus i tried to find a more appropriate concept scheme but no luck, should I make a new concept scheme?

@Windvis i tried to make the forms readonly but no luck... what should I pass to rdfaForm?

Windvis:
> @Windvis i tried to make the forms readonly but no luck... what should I pass to rdfaForm?

@Asergey91  I think you can simply set the `@show` argument of the RdfForm component based on the status value? 

Similar to how it's done in the ember-submission-form-fields test app: https://github.com/lblod/ember-submission-form-fields/blob/9c4cd929ac6ecdee08b0f0fbba4014180e4d7e7f/tests/dummy/app/components/test-form.hbs#L7

Asergey91:
> > @Windvis i tried to make the forms readonly but no luck... what should I pass to rdfaForm?
> 
> @Asergey91 I think you can simply set the `@show` argument of the RdfForm component based on the status value?
> 
> Similar to how it's done in the ember-submission-form-fields test app: https://github.com/lblod/ember-submission-form-fields/blob/9c4cd929ac6ecdee08b0f0fbba4014180e4d7e7f/tests/dummy/app/components/test-form.hbs#L7

ah but that breaks the rdfaEditor

Asergey91:
https://github.com/lblod/ember-submission-form-fields/pull/98

app/components/rdf-form-fields/rich-text-editor.hbs
```diff
@@ -11,10 +11,12 @@
 {{/if}}
 
 {{#if @show}}
-  <div class="rich-text-editor-content">
-    {{! template-lint-disable no-triple-curlies}}
-    {{{this.value}}}
-  </div>
+  {{#if this.value}}
```
Windvis, line 14: This will hide the visual "content" and only show the label which looks a bit weird to me. Why is this check needed exactly? If it's something quirky with `{{{this.value}}}` I would only conditionally render that so we show at least an empty looking "input"?

If the "empty input" looks weird as well we could maybe use the `<ValueWithPlaceholder>` component to show a `-` if there is no value.

```hbs
          <ValueWithPlaceholder @value={{this.value}}>
            <:value as |value|>
              <div class="rich-text-editor-content">
                {{! template-lint-disable no-triple-curlies}}
                {{{value}}}
             </div>
            </:value>
          </ValueWithPlaceholder>
```

app/utils/constants.js
```diff
@@ -0,0 +1,16 @@
+export const FORM_MAPPING = {
+  'cd0b5eba-33c1-45d9-aed9-75194c3728d3': 'inhoud',
```
Windvis, line 2: Also this config is highly specific to the component, so I wouldn't move it out of there. The generic config that _could_ be extracted is the uuid of the form. We could in theory reuse these ids in the tab routes then but for now I would simply keep it in the component as before.

app/components/public-services/details-page.js
```diff
@@ -90,6 +85,11 @@ export default class PublicServicesDetailsPageComponent extends Component {
     this.hasUnsavedChanges = true;
   }
 
+  get readOnly() {
+    const statusUri = this.args.publicService.status.get('uri')
+    return statusUri !== LIFECYCLE_STATUS.CONCEPT;
```
Windvis, line 90: `LIFECYCLE_STATUS` doesn't exist anymore so I think this throws?

Windvis (changes requested):

app/components/rdf-form-fields/rich-text-editor.hbs
Asergey91: it looked wierd to me imho, but i dont mind changing it back

Windvis: Yea, agreed. This is probably just something that Dieter needs to look at.

Windvis (approved):

PR #165, opened 2022-08-23
Add a custom `ConceptSelector` field

Windvis:
This field works similar to the `ConceptSchemeSelector` and `ConceptSchemeMultiSelector` fields, but with several improvements:

- it loads the data when the component is rendered or when the user searches, which improves the form rendering
- it allows preloading the options in case there aren't that many so searching isn't beneficial
- it supports both single and multi select modes by adding some configuration

It can be tested in the LPDC module by adjusting `conceptSchemeSelector` and `conceptSchemeMultiSelector` display types to `conceptSelector` and providing options through `form:options`.

Available options:
```ts
type Options = {
  conceptScheme: string; // required and sets the concept scheme to which the concepts belong
  multiple?: boolean; // switches to the multi-select mode if set to true. defaults to `false`
  preload?: boolean; // loads some records when the component is rendered. defaults to `true`
  preloadAmount?: int; // the amount of records that will be preloaded, defaults to `20`.
  search?: boolean; // should we show the search field in the select component. defaults to `true`
  backendSearch?: boolean; // should we request data from the backend when searching. If set to false it will only search in the preloaded options. This option requires that `search` is set to `true`. defaults to `true`
}
```

### DIFF
--- a/app/components/rdf-form-fields/concept-selector.hbs
+++ b/app/components/rdf-form-fields/concept-selector.hbs
@@ -9,8 +9,8 @@
   <this.selectComponent
     @triggerId={{this.inputId}}
     @selected={{this.selected}}
-    @searchEnabled={{not this.shouldPreloadData}}
-    @search={{perform this.search}}
+    @searchEnabled={{this.isSearchEnabled}}
+    @search={{if this.backendSearch (perform this.search)}}
     @searchField="label"
     @options={{this.options}}
     @onClose={{fn (mut this.hasBeenFocused) true}}
--- a/app/components/rdf-form-fields/concept-selector.hbs
+++ b/app/components/rdf-form-fields/concept-selector.hbs
@@ -1,0 +1,32 @@
+<AuLabel
+  @error={{this.hasErrors}}
+  @required={{this.isRequired}}
+  for={{this.inputId}}
+>
+  {{@field.label}}
+</AuLabel>
+<div class={{if this.hasErrors "ember-power-select--error"}}>
+  <this.selectComponent
+    @triggerId={{this.inputId}}
+    @selected={{this.selected}}
+    @searchEnabled={{not this.shouldPreloadData}}
+    @search={{perform this.search}}
+    @searchField="label"
+    @options={{this.options}}
+    @onClose={{fn (mut this.hasBeenFocused) true}}
+    @onChange={{this.updateSelection}}
+    @allowClear={{true}}
+    @searchMessage={{this.searchMessage}}
+    @loadingMessage="Aan het laden..."
+    @noMatchesMessage="Geen resultaten gevonden"
+    @disabled={{@show}}
+    data-test-field-uri={{@field.uri.value}}
+    as |concept|
+  >
+    {{concept.label}}
+  </this.selectComponent>
+</div>
+
+{{#each this.errors as |error|}}
+  <AuHelpText @error={{true}}>{{error.resultMessage}}</AuHelpText>
+{{/each}}
--- a/app/components/rdf-form-fields/concept-selector.js
+++ b/app/components/rdf-form-fields/concept-selector.js
@@ -1,0 +1,132 @@
+import { action } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import InputFieldComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/input-field';
+import {
+  triplesForPath,
+  updateSimpleFormValue,
+  removeDatasetForSimpleFormValue,
+} from '@lblod/submission-form-helpers';
+import { restartableTask, timeout } from 'ember-concurrency';
+import { NamedNode } from 'rdflib';
+import PowerSelect from 'ember-power-select/components/power-select';
+import PowerSelectMultiple from 'ember-power-select/components/power-select-multiple';
+
+export default class RdfFormFieldsConceptSchemeSelectorComponent extends InputFieldComponent {
+  @service store;
+
+  @tracked selected = null;
+  options = [];
+  inputId = 'concept-selector-' + guidFor(this);
+
+  constructor() {
+    super(...arguments);
+    this.loadSelectOptions();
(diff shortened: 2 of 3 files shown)


app/components/rdf-form-fields/concept-selector.js
```diff
@@ -0,0 +1,116 @@
...
+    return Boolean(this.args.field.options.multiple);
+  }
+
+  get selectComponent() {
+    return this.multiple ? PowerSelectMultiple : PowerSelect;
+  }
+
+  loadOptions() {
+    if (this.shouldPreloadData) {
+      this.options = this.loadConcepts();
```
Windvis, line 47: At the moment this only preloads the 20 first records, or more if `options.preloadAmount` is set in the form config. I think that covers the use case where we only want use the preload mode if the amount if the amount of options is limited.

If that assumption was wrong we could also make it fetch all the records? In that case we probably want to allow searching in the options as well.

app/components/rdf-form-fields/concept-selector.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  loadConcepts(query = {}) {
+    let { conceptScheme, preloadAmount = 20 } = this.args.field.options;
+
+    return this.store.query('concept', {
```
Windvis, line 54: Using Ember Data here isn't really a generic solution since it requires the modal classes to be defined exactly the same as the `concept` and `concept-scheme` classes in Loket. If we ever want to move this component to the addon we probably need a different solution.

app/components/rdf-form-fields/concept-selector.js
```diff
@@ -0,0 +1,116 @@
...
+  inputId = 'concept-selector-' + guidFor(this);
+
+  constructor() {
+    super(...arguments);
+    this.loadOptions();
+    this.loadProvidedValue();
+  }
+
+  get searchMessage() {
+    return this.args.field.options?.searchMessage || 'Typ om te zoeken';
```
Windvis, line 30: This can be used to provide a better message depending on the context, for example "Typ om een overheid te zoeken". We could also use it to guide users with examples, but the help text might be better suited for that.

app/components/rdf-form-fields/concept-selector.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  async loadProvidedValue() {
+    const matches = triplesForPath(this.storeOptions, true).values;
+
+    if (matches.length > 0) {
+      let conceptUri = matches[0].value;
+      let concepts = await this.store.query('concept', {
+        'filter[:uri:]': conceptUri,
+      });
```
Windvis, line 69: This still needs to be adjusted for the `multiple` mode.

app/components/rdf-form-fields/concept-selector.js
cecemel: check

app/components/rdf-form-fields/concept-selector.js
```diff
@@ -0,0 +1,116 @@
...
+  get searchMessage() {
+    return this.args.field.options?.searchMessage || 'Typ om te zoeken';
+  }
+
+  get shouldPreloadData() {
+    return Boolean(this.args.field.options?.preload);
+  }
+
+  get multiple() {
+    return Boolean(this.args.field.options.multiple);
```
cecemel, line 38: inconsistency with https://github.com/lblod/frontend-loket/pull/165/files#diff-df4e772c616d1e06578859960a5ac29f602084aed33cd80ef58ccbd157723bc8R34

the ? operator

app/components/rdf-form-fields/concept-selector.js
cecemel: the preload everything could be an extension? I think we can start grinding with current  behaviour

app/components/rdf-form-fields/concept-selector.js
cecemel: yeah agree. I don't know when we will hit the following step first:
- it becomes a sparql implementation
- we need other things to load. In that case a lot of things have to change; and even the name wouldn't make sense anymore and skos:inScheme filter neither. 

So to me, currently fine with this.  A lot of use cases will be covered with current implementation. (In the past all the reasons why we have been sub-classing concepts, it was because we didn't have  inheritance).

app/components/rdf-form-fields/concept-selector.js
cecemel: check

cecemel (commented):
Code seems fine, bar nitpick. 
Still need to test, later today.

app/components/rdf-form-fields/concept-selector.js
Windvis: Right, I wanted to ask about that. The component _needs_ a concept scheme uri to function properly, which means the options field should always be set. I think we can drop al the optional chaining.

It might even be beneficial to throw an assert if the conceptScheme isn't set?

Windvis: I've dropped the optional chaining since it defaults to an empty object in the field model class.

app/components/rdf-form-fields/concept-selector.hbs
```diff
@@ -0,0 +1,32 @@
...
+  @required={{this.isRequired}}
+  for={{this.inputId}}
+>
+  {{@field.label}}
+</AuLabel>
+<div class={{if this.hasErrors "ember-power-select--error"}}>
+  <this.selectComponent
+    @triggerId={{this.inputId}}
+    @selected={{this.selected}}
+    @searchEnabled={{not this.shouldPreloadData}}
```
cecemel, line 12: not sure if this is desired. Sometimes we want the user to get inspiration to look for words

cecemel (changes requested):

app/components/rdf-form-fields/concept-selector.hbs
Windvis: I've updated the config options. It is now possible to:
- also preload records even if we search on the backend.
- search client side only, in case we preloaded all the data but there are still a lot of them.

(branch reset is needed because I force pushed)

PR #166, opened 2022-08-24
Attempt at toggling icons in the overview table when public service is sent.

cecemel:


app/models/public-service.js
```diff
@@ -4,4 +4,8 @@ import ConceptualPublicServiceModel from './conceptual-public-service';
 export default class PublicServiceModel extends ConceptualPublicServiceModel {
   @belongsTo('conceptual-public-service', { inverse: null })
   concept;
+
+  get isSent() {
```
cecemel, line 8: is this ok?

Windvis: It will work now, but accessing async relationships without `await`ing them first is going to be [deprecated soon](https://rfcs.emberjs.com/id/0745-ember-data-deprecate-methods-on-promise-many-array). I wouldn't depend on it anymore to be honest. This is one of the reasons why I want to switch to sync relationships, things like this become much easier to reason about since the async accessing is gone.

We could probably use a helper instead? Since it's now possible to use plain functions as helpers, I would export a new function from the model class which we import where needed.

(These are just examples, naming can be tweaked 😄)

```js
// app/models/public-service.js
export function isSent(statusUri) {
  return statusUri === 'http://lblod.data.gift/concepts/9bd8d86d-bb10-4456-a84e-91e9507c374c';
}
```

And then in the controller of the route we could import this function and store it on a property and access it in the template:

```js
// app/controllers/public-services/index.js
import { isSent } from 'frontend-loket/models/public-service';

export default class PublicServicesIndexController extends Controller {
  //...
  isSent = isSent
```

```hbs
{{! app/templates/public-services/index.hbs}}
{{#if (this.isSent publicService.status.uri)}}
...
{{/if}}
```

Alternatively we could make it a global helper
```js
// app/helpers/is-public-service-sent.js
export function isPublicServiceSent(statusUri) {
  return statusUri === 'http://lblod.data.gift/concepts/9bd8d86d-bb10-4456-a84e-91e9507c374c';
}
```

```hbs
{{! app/templates/public-services/index.hbs}}
{{#if (is-public-service-sent publicService.status.uri)}}
...
{{/if}}
```

cecemel: ok, thanks for the very extended explanation. Good to know we already can use js functions now as helpers. And providing a pattern to do so. 
Anyway, the async false seemed to the most transparent solution, so let's try this

app/templates/public-services/index.hbs
```diff
@@ -134,15 +134,27 @@
...
-            @iconAlignment="left"
-            @skin="primary"
-            @route="public-services.details"
-            @model={{publicService.id}}
-          >
-            Bewerk
-          </AuLink>
+          {{#if publicService.isSent}}
+            <AuLink
+              @icon="eye"
```
Windvis, line 139: I would set the icon and button label conditionally instead of duplicating the whole block.

```suggestion
              @icon={{if publicService.isSent "eye" "pencil"}}
```

app/templates/public-services/index.hbs
```diff
@@ -134,15 +134,27 @@
...
+              @iconAlignment="left"
+              @skin="primary"
+              @route="public-services.details"
+              @model={{publicService.id}}
+            >
+              Bekijk
```
Windvis, line 145: Same here

```suggestion
              {{if publicService.isSent "Bekijk" "Bewerk"}}
```

Windvis (approved):

PR #167, opened 2022-08-24
Chore/some corrections lpdc details page

cecemel:
Some corrections, it is still in draft, but i have no time to pick it further up.  Any candidates @Asergey91 @Windvis 
- I think it needs more try/catches
- It think component state management is a bit over the place. should be fine tuned
  - this.sending
- this.args is a bit over the place, better in a getter?

I tried bundling the improvements per commit, so it is more clear what it does

### DIFF
--- a/app/components/public-services/details-page.js
+++ b/app/components/public-services/details-page.js
@@ -10,6 +10,7 @@ import {
 import rdflib from 'rdflib';
 import { dropTask, task, dropTaskGroup } from 'ember-concurrency';
 import ConfirmDeletionModal from 'frontend-loket/components/public-services/confirm-deletion-modal';
+import ConfirmSubmitModal from 'frontend-loket/components/public-services/confirm-submit-modal';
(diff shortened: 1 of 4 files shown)


Windvis:
- I've moved the try catches into a separate PR (which targets this one): #170 
- I replaced the modal state with the modal service we already used for the delete and unsaved changes modals.
- unchanged for now as discussed in chat

app/components/public-services/details-page.js
```diff
@@ -168,9 +165,15 @@ export default class PublicServicesDetailsPageComponent extends Component {
     );
 
     yield loadPublicServiceDetails(this.store, this.args.publicService.id);
-    this.hasUnsavedChanges = false;
+  }
 
-    this.submitToGoverment = false;
+  @action
+  requestSubmitConfirmation() {
+    this.modals.open(ConfirmSubmitModal, {
```
Windvis, line 172: We could validate the current form before opening the modal, that might give faster feedback to the user which I think was one of the user testing remarks?

cecemel: yes!

PR #171, opened 2022-08-24
Replace hardcoded RDF namespaces with imports

Windvis:
We were defining a lot of custom rdflib namespaces which were either already exported, or used in multiple places.

This refactors the code a bit to reuse the existing ones, and make the others more reusable.

### DIFF
--- a/app/components/public-services/details-page.js
+++ b/app/components/public-services/details-page.js
@@ -7,12 +7,12 @@ import {
   ForkingStore,
   validateForm,
 } from '@lblod/ember-submission-form-fields';
-import { FORM, RDF } from '@lblod/submission-form-helpers';
 import { NamedNode } from 'rdflib';
 import { dropTask, task, dropTaskGroup } from 'ember-concurrency';
 import ConfirmDeletionModal from 'frontend-loket/components/public-services/confirm-deletion-modal';
 import ConfirmSubmitModal from 'frontend-loket/components/public-services/confirm-submit-modal';
 import UnsavedChangesModal from 'frontend-loket/components/public-services/details/unsaved-changes-modal';
+import { FORM, RDF } from 'frontend-loket/rdf/namespaces';
 import { loadPublicServiceDetails } from 'frontend-loket/utils/public-services';
 
 const FORM_MAPPING = {
--- a/app/components/public-services/details-page.js
+++ b/app/components/public-services/details-page.js
@@ -7,7 +7,8 @@ import {
   ForkingStore,
   validateForm,
 } from '@lblod/ember-submission-form-fields';
-import { NamedNode, Namespace } from 'rdflib';
+import { FORM, RDF } from '@lblod/submission-form-helpers';
+import { NamedNode } from 'rdflib';
 import { dropTask, task, dropTaskGroup } from 'ember-concurrency';
 import ConfirmDeletionModal from 'frontend-loket/components/public-services/confirm-deletion-modal';
 import ConfirmSubmitModal from 'frontend-loket/components/public-services/confirm-submit-modal';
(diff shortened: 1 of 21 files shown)


app/rdf/namespaces.js
```diff
@@ -0,0 +1,8 @@
+import { Namespace } from 'rdflib';
+
+export const DBPEDIA = new Namespace('http://dbpedia.org/ontology/');
+export const LBLOD_SUBSIDIE = new Namespace(
+  'http://lblod.data.gift/vocabularies/subsidie/'
+);
+export const MU = new Namespace('http://mu.semte.ch/vocabularies/core/');
```
Windvis, line 7: The MU namespace [already exists](https://github.com/lblod/submission-form-helpers/blob/7b57b1da18414baf47e331e22fee290835465b4b/src/namespaces.js#L10) in `@lblod/submission-form-helpers` but [it isn't exported](https://github.com/lblod/submission-form-helpers/blob/7b57b1da18414baf47e331e22fee290835465b4b/src/constraints.js#L165) from the entry point so we can't import it here.

app/components/rdf-form-fields/estimated-cost-table/base-table.js
```diff
@@ -2,9 +2,7 @@ import { tracked } from '@glimmer/tracking';
 
 import InputFieldComponent from '@lblod/ember-submission-form-fields/components/rdf-input-fields/input-field';
 import { triplesForPath } from '@lblod/submission-form-helpers';
-import { NamedNode, Namespace } from 'rdflib';
-
-export const MU = new Namespace('http://mu.semte.ch/vocabularies/core/');
```
Windvis, line 7: This wasn't used internally and I replaced the usage in the subclass with the generic import.

app/components/public-services/details-page.js
```diff
@@ -7,6 +7,7 @@ import {
   ForkingStore,
   validateForm,
 } from '@lblod/ember-submission-form-fields';
+import { FORM, RDF } from '@lblod/submission-form-helpers';
```
Windvis, line 10: It might be better to also import these from the `app/rdf/namespaces.js` file since that allows us to manage them in a single place.

app/rdf/namespaces.js
cecemel: might be handy yes

app/components/rdf-form-fields/estimated-cost-table/base-table.js
cecemel: ok

app/components/public-services/details-page.js
cecemel: yes

app/components/rdf-form-fields/climate-subsidy-costs-table/table-row-burgemeesters.js
```diff
@@ -3,11 +3,10 @@ import { tracked } from '@glimmer/tracking';
...
 import { A } from '@ember/array';
 import { scheduleOnce } from '@ember/runloop';
-import { literal, NamedNode, Namespace } from 'rdflib';
+import { literal, NamedNode } from 'rdflib';
 import { v4 as uuidv4 } from 'uuid';
-import { RDF, XSD } from '@lblod/submission-form-helpers';
-
-const MU = new Namespace('http://mu.semte.ch/vocabularies/core/');
+import { RDF, XSD } from 'frontend-loket/rdf/namespaces';
+import { MU } from 'frontend-loket/rdf/namespaces';
```
Windvis, line 9: Tthese can be combined into one import.

app/rdf/namespaces.js
Windvis: [PRd in `@lblod/submission-form-helpers`](https://github.com/lblod/submission-form-helpers/pull/28). We can swap it for the import when we update to a future version.

PR #172, opened 2022-08-25
Update to Appuniversum v1.4.1+

Windvis:
~~Wrapping the app in `AuApp` solves a lot of issues, but not everything. The body container previously used display flex + columns, which we depended on in some pages to align toolbars to the bottom (while making the content above scrollable). This no longer works and that can be seen in the Toezicht module.~~

The issue was fixed in Appuniversum v1.4.1 so we can unpin the version. It also includes the `AuApp` wrapper since that's a good change regardless and I had already written the code.

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,98 +1,110 @@
 {{page-title this.appTitle}}
-<AuToaster />
-<AuMainHeader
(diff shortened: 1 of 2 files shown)


Windvis:
This issue has been fixed in v1.4.1, but it's still a good idea to wrap the app in a AuApp component. We should unpin the version lock and check if everything works.

PR #176, opened 2022-08-30
Adding nationality field in persons component

Poltergeistz:
DL-4123

### DIFF
--- a/app/components/shared/persoon/create-persoon.hbs
+++ b/app/components/shared/persoon/create-persoon.hbs
@@ -87,6 +87,29 @@
           {{/if}}
         {{/let}}
       </div>
+      {{#if this.isNationalityFieldRequired}}
+      {{#let (if this.errors.nationaliteit true false) as |showError|}}
+        <div>
+          <AuLabel for="mandataris-nationaliteit"
+            @error={{showError}}
+            @required={{true}}
+            >
+            Nationaliteit
+          </AuLabel>
+
+          <Shared::Persoon::NationalitySelector
+            @id="mandataris-nationaliteit"
+            @selected={{this.nationaliteit}}
+            @onChange={{fn (mut this.nationaliteit)}}
+          />
+        </div>
+        {{#if showError}}
+          <AuHelpText @error="true">
+            {{this.errors.nationaliteit}}
+          </AuHelpText>
+        {{/if}}
+      {{/let}}
+      {{/if}}
       <div>
         <AuLabel for="mandataris-geboorte">Geboortedatum</AuLabel>
         <AuDatePicker
--- a/app/components/shared/persoon/create-persoon.hbs
+++ b/app/components/shared/persoon/create-persoon.hbs
@@ -99,8 +99,10 @@
 
           <Shared::Persoon::NationalitySelector
             @id="mandataris-nationaliteit"
+            @allowClear={{not this.isNationalityFieldRequired}}
             @selected={{this.nationaliteit}}
             @onChange={{fn (mut this.nationaliteit)}}
+            @error={{showError}}
           />
         </div>
         {{#if showError}}
--- a/app/components/shared/persoon/create-persoon.hbs
+++ b/app/components/shared/persoon/create-persoon.hbs
@@ -90,7 +90,7 @@
       {{#if this.isNationalityFieldRequired}}
       {{#let (if this.errors.nationaliteit true false) as |showError|}}
         <div>
-          <AuLabel for="mandataris-nationaliteit"
+          <AuLabel for="person-nationalities"
             @error={{showError}}
             @required={{true}}
             >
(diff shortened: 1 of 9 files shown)


app/components/shared/persoon/nationality-selector.hbs
```diff
@@ -0,0 +1,16 @@
+<PowerSelect
+  @allowClear={{true}}
```
Windvis, line 2: I would make this an argument since it makes sense to disable it if selecting a nationality is required.

app/models/persoon.js
```diff
@@ -7,6 +7,7 @@ export default class PersoonModel extends Model {
   @belongsTo('geboorte', { inverse: null }) geboorte;
   @belongsTo('identificator', { inverse: null }) identificator;
   @belongsTo('geslacht-code', { inverse: null }) geslacht;
+  @belongsTo('nationality') nationality;
```
Windvis, line 10: Explicit inverses are recommended since they will be a requirement in the future.
```suggestion
  @belongsTo('nationality', { inverse: null }) nationality;
```

app/components/shared/persoon/nationality-selector.js
```diff
@@ -0,0 +1,22 @@
...
+
+    const query = {
+      sort: 'nationality-label',
+    };
+
+    if (search.trim() !== '') {
+      query['filter[nationality-label]'] = search;
+    }
+
+    return yield this.store.query('nationality', query);
```
Windvis, line 20: This will also search even if the user didn't type a letter. I think we simply want to return an empty array in that case?

app/components/shared/persoon/nationality-selector.hbs
```diff
@@ -0,0 +1,16 @@
+<PowerSelect
```
Windvis, line 1: I would wrap this in an extra div with conditional error classes so the select itself also receives a red border if nothing is selected.

Similar to how we do it here: https://github.com/lblod/frontend-loket/blob/c6ed2cd7fe28e450fdb0f343400e8abe8e2926be/app/components/worship-ministers-management/position-select.hbs#L1

(minus the `data-test` attribute)

Windvis (commented):
Code looks good, only some small remarks 👍.

app/components/shared/persoon/nationality-selector.js
Poltergeistz: Made the change with a RegExp, I don't know if it's efficient as it does not take accents but characters that are not letters are now excluded.
```
if (/^[a-zA-Z]+$/.test(search.trim())) {
      query['filter[nationality-label]'] = search;
      return yield this.store.query('nationality', query);
    }

    return [];
```

Windvis: That's not what I meant, sorry 😅. I didn't want to limit the search to letters only, but to input only. It seemed strange to do a search with an "empty" input since it would simply return the first page of nationalities. It seemed more logical to me to return an empty array in that case.

I don't think actually restricting it to only letters is correct either since there might be countries with non letter characters (spaces, `-`, ...). So I would revert to the previous empty check.

app/models/persoon.js
Windvis: Oh, right, this also needs to be a hasMany relationships since a person can have multiple nationalities.

app/components/shared/persoon/nationality-selector.hbs
Windvis: This should also be a `PowerSelectMultiple` since a person can have more than 1 nationality.

app/models/persoon.js
Poltergeistz: Okay, I'm doing changes for the backend. There will be some modifications here too with the relationship changes.

app/components/shared/persoon/create-persoon.hbs
```diff
@@ -87,6 +87,31 @@
           {{/if}}
         {{/let}}
       </div>
+      {{#if this.isNationalityFieldRequired}}
+      {{#let (if this.errors.nationaliteit true false) as |showError|}}
+        <div>
+          <AuLabel for="mandataris-nationaliteit"
```
Windvis, line 93: Woops, missed this. This component is used in multiple places so I would use a more generic id.
```suggestion
          <AuLabel for="person-nationalities"
```

app/components/shared/persoon/create-persoon.hbs
```diff
@@ -87,6 +87,31 @@
...
+            @error={{showError}}
+            @required={{true}}
+            >
+            Nationaliteit
+          </AuLabel>
+
+          <Shared::Persoon::NationalitySelector
+            @id="mandataris-nationaliteit"
```
Windvis, line 101: (same here)

Windvis: Also, I think this needs to be an attribute. At the moment the id is not passed to the power select component.
```suggestion
            id="person-nationalities"
```

Windvis (commented):
Last round, sorry 😄.

Windvis (approved):

PR #179, opened 2022-09-01
Changing `contacts` validation

Poltergeistz:
DL-4124

This sets the fields for address, primary phone number and e-mail to required. Also changes the error message.

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -77,7 +77,7 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       let secondaryContactPoint = yield contactPoint.secondaryContactPoint;
       let adres = yield contactPoint.adres;
 
-      if (contactPoint.telefoon || contactPoint.email || adres) {
+      if (contactPoint.telefoon && contactPoint.email && adres) {
         if (secondaryContactPoint.telefoon) {
           yield secondaryContactPoint.save();
         } else {
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -7,6 +7,7 @@ import {
   createPrimaryContactPoint,
   createSecondaryContactPoint,
   findPrimaryContactPoint,
+  isValidPrimaryContact,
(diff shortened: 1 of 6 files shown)


app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -77,7 +77,7 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       let secondaryContactPoint = yield contactPoint.secondaryContactPoint;
       let adres = yield contactPoint.adres;
 
-      if (contactPoint.telefoon || contactPoint.email || adres) {
```
Windvis, line 80: I think the current validation system can be improved a bit.

At the moment we add an [error for every required field](https://github.com/lblod/frontend-loket/blob/2530d959b90d67e3f2ebd4ee4f9c643a5eedf8c7/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js#L101-L104). This was fine before, but now that they are all required I would only add errors for the ones that are actually empty.

I would create a new `isValidPrimaryContact` utility function that receives a contactPoint record, checks the values and adds the errors where needed. This also has the benefit that it can be reused in the 2 modules. It could return a boolean which we can use to check if the record is valid, which is easy to check with the record's [`isValid` property](https://api.emberjs.com/ember-data/4.6/classes/Model/properties/isValid?anchor=isValid).

Another change that's needed is in the contact-information/row component. It seems I didn't fully implement the error state for the fields. [`.error`](https://github.com/lblod/frontend-loket/blob/2530d959b90d67e3f2ebd4ee4f9c643a5eedf8c7/app/components/contact-information-table/row.hbs#L25) should become `errors` for the email and primary phone fields. That should show the red borders if there are any errors.

[The address field](https://github.com/lblod/frontend-loket/blob/2530d959b90d67e3f2ebd4ee4f9c643a5eedf8c7/app/components/contact-information-table/row.hbs#L9-L11) should also show a border, so I [would wrap it in a div](url) with a conditional error class similar with the other power-select wrappers.

A last thing that would need a change is the error alert. It should only be displayed if the user is editing a contact, _and_ if that contact is invalid. So [this check](https://github.com/lblod/frontend-loket/blob/2530d959b90d67e3f2ebd4ee4f9c643a5eedf8c7/app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs#L99-L105) needs to be tweaked.

PR #182, opened 2022-09-02
Added climate subsidy version (LEKP1.0 or LEKP2.0)  differentiation o…

cecemel:
…ption

### DIFF
--- a/app/components/rdf-form-fields/climate-subsidy-costs-table.js
+++ b/app/components/rdf-form-fields/climate-subsidy-costs-table.js
@@ -136,11 +136,9 @@ export default class RdfFormFieldsClimateSubsidyCostsTableComponent extends Inpu
       metaGraph
     )[0].object.value;
 
-
-
     this.drawingRight = drawingRight;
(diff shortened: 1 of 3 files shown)


app/components/rdf-form-fields/climate-subsidy-costs-table.js
```diff
@@ -128,8 +129,16 @@ export default class RdfFormFieldsClimateSubsidyCostsTableComponent extends Inpu
       metaGraph
     )[0].object.value;
 
+    const lekpValidation = this.args.formStore.match(
+      undefined,
+      LBLOD_SUBSIDIE('lekp'),
+      undefined,
+      metaGraph
+    )[0].object.value;
```
Windvis, line 137: Just to be sure, the v1 version also has this predicate set (but to `1.0`)?

cecemel: eagle eyes, but yes, they should be there :-)

PR #183, opened 2022-09-05
Changing default heading text

Poltergeistz:
DL-4119

Do we keep the default button text ?

### DIFF
--- a/app/templates/eredienst-mandatenbeheer/new.hbs
+++ b/app/templates/eredienst-mandatenbeheer/new.hbs
@@ -1,21 +1,10 @@
 {{page-title "Voeg mandaat toe"}}
 
-{{#if this.shouldSelectPerson}}
-  <Shared::Persoon::PersoonSearchForm
-    class="au-c-body-container"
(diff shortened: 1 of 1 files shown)


app/templates/eredienst-mandatenbeheer/new.hbs
```diff
@@ -6,8 +6,9 @@
     @onCreateNewPerson={{this.createNewPerson}}
     @onCancel={{this.cancel}}
     @onSelect={{this.selectPersoon}}
+    @showDefaultHead={{true}}
```
Windvis, line 9: We actually don't want the default header. We want to disable it so we can add a new one above this component.

So this should be set to false, and the block content should be removed.

It should just mimic this setup we use in the other module: https://github.com/lblod/frontend-loket/blob/e158a465889f0009bcda2c05a0bf7a7d16b68f7b/app/templates/worship-ministers-management/new.hbs#L3-L45

Windvis (approved):

PR #186, opened 2022-09-07
LPDC save improvements

Windvis:
- prevent the unsaved changes modal from showing after publishing
- update the modified date after saving


cecemel:
Even though it was merged, it works :-)

PR #187, opened 2022-09-07
Fix duplicate entries in the mandate position select

Poltergeistz:
DL-4163

We get all IDs when not selecting a bestuursperiode, leading to duplicate results when we call `mandaat?filter[bevat-in][id]={IDs}&sort=bestuursfunctie.label`.

This changes the order of entities inside the `BestuursperiodenSelector` so the recent date is set first.

When a user isn't selecting a bestuursperiode and wants to create a new mandatee, the bestuursperiode by default is the most recent on the `new` route.

### DIFF
--- a/app/routes/eredienst-mandatenbeheer/new.js
+++ b/app/routes/eredienst-mandatenbeheer/new.js
@@ -23,7 +23,6 @@ export default class EredienstMandatenbeheerNewRoute extends Route {
         this.store.findRecord('persoon', personId, {
           backgroundReload: false,
         }),
-        this.store.findAll('half-election'),
         this.bestuursorganen.length > 1
           ? [this.bestuursorganen.at(0)] // The recent bindingStart selected by default
           : this.bestuursorganen,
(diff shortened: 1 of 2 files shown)


app/routes/eredienst-mandatenbeheer/new.js
```diff
@@ -22,7 +23,10 @@ export default class EredienstMandatenbeheerNewRoute extends Route {
         this.store.findRecord('persoon', personId, {
           backgroundReload: false,
         }),
-        this.bestuursorganen,
+        this.store.findAll('half-election'),
```
claire-lovisa, line 26: The half-elections got removed in #184 , this is breaking the flow no ?

claire-lovisa (changes requested):
Removing the `reverse()` impacts the behaviour of the `bestuursperioden-selector` on other pages such as the Mandatenbeheer of non-worship admin units, which often have periods in the past. I feel like it's weird that we have the first period it time as a default one.

removing-reverse-impacts-other-modules

@cecemel what's your opinion ? It's indeed better for worship services but makes it weird for the other unit types. Should we just keep the reverse and hide organen that are in the future ?

cecemel:
@claire-lovisa @Poltergeistz 
I the less confusing would be, i.e. default behaviour (so if none selected)
- select the bestuursorgaan matching the period where we are in today
-  if none is found, take the next starting one in the future?
-?

cecemel:
Closing because fixed in #196

PR #188, opened 2022-09-08
Add a `TagSelector` form field component

Windvis:
Depends on https://github.com/lblod/app-digitaal-loket/pull/228

At the moment the user can only add new tags. Searching for and adding already existing ones will be implemented in the future.

### DIFF
--- a/app/components/public-services/rdf-form-fields/tag-selector.hbs
+++ b/app/components/public-services/rdf-form-fields/tag-selector.hbs
@@ -6,25 +6,40 @@
   {{@field.label}}
 </AuLabel>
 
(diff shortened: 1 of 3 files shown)


Windvis:
The disabled state looks a bit strange and probably needs to be fixed in Appuniversum:

image

Alternatively we could use a list of [`AuPill`](https://appuniversum.github.io/ember-appuniversum/?path=/story/components-notifications-aupill--component) components instead of disabling the select component.

cecemel:
cool, working cool. perhaps we could do the pills to for everything that is multiselect?

Windvis:
> cool, working cool. perhaps we could do the pills to for everything that is multiselect?

I'll create an issue for it 👍

PR #190, opened 2022-09-12
Allow users to reopen a submitted service

Windvis:
It works, but the copy for the buttons and labels still needs to be changed.

npm install is needed since I installed the refresh polyfill.

### DIFF
--- a/app/components/public-services/confirm-reopening-modal.hbs
+++ b/app/components/public-services/confirm-reopening-modal.hbs
@@ -5,10 +5,8 @@
 >
   <:title>Product of dienst opnieuw bewerken?</:title>
   <:body>
-    {{! TODO: add proper copy }}
-    Om wijzigingen aan te brengen moet het product eerst naar de ontwerp status
-    worden gebracht. Nieuwe wijzigingen zullen daarna opnieuw gepubliceerd
-    moeten worden.
+  Eenmaal heropend, komt het document terug in status 'Ontwerp' te staan. <br>
+  Nadien kan u verder werken zoals gewoonlijk.
   </:body>
   <:footer>
     <AuButtonGroup>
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
     "ember-promise-modals": "^3.0.1",
     "ember-qunit": "^5.1.5",
     "ember-resolver": "^8.0.3",
+    "ember-router-service-refresh-polyfill": "^1.0.1",
```
Windvis, line 86: We can remove this again once we update to Ember 4.4.

app/components/public-services/confirm-reopening-modal.hbs
```diff
@@ -0,0 +1,35 @@
...
+  @modalOpen={{true}}
+  @closeModal={{this.close}}
+  data-test-confirm-reopening-modal="modal"
+>
+  <:title>Product of dienst opnieuw bewerken?</:title>
+  <:body>
+    {{! TODO: add proper copy }}
+    Om wijzigingen aan te brengen moet het product of dienst opnieuw naar de
+    ontwerp status worden gebracht. Nieuwe wijzigingen moeten daarna opnieuw
+    gepubliceerd worden voor deze extern zichtbaar zijn.
```
Windvis, line 11: I just added some quick copy, but it doesn't make much sense and I didn't want to lose too much time on it.

@cecemel suggestions? Same for the title and buttons.

app/components/public-services/confirm-reopening-modal.hbs
```diff
@@ -0,0 +1,34 @@
...
+  @closeModal={{this.close}}
+  data-test-confirm-reopening-modal="modal"
+>
+  <:title>Product of dienst opnieuw bewerken?</:title>
+  <:body>
+    {{! TODO: add proper copy }}
+    Om wijzigingen aan te brengen moet het product eerst naar de ontwerp status
+    worden gebracht. Nieuwe wijzigingen zullen daarna opnieuw gepubliceerd
+    moeten worden.
+  </:body>
```
Windvis, line 12: We need better copy, or we can only show the title, but I'm not sure if that is clear enough.

PR #192, opened 2022-09-14
Set up release-it

Windvis:
This replaces ember-cli-release with release-it. ember-cli-release has been deprecated for a while and it brought in an old npm version which sometimes caused issues.

`npm run release` is all that's needed to create a new release. release-it will then prompt for some basic information.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,5 @@
+{
+  "npm": {
+    "publish": false
+  }
+}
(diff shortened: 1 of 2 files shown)


.release-it.json
```diff
@@ -0,0 +1,5 @@
+{
+  "npm": {
+    "publish": false
```
Windvis, line 3: This is only needed because release-it still shows the "publish to npm" prompt, even though `private: true` is set in the package.json. Once that is resolved we can remove the config file and depend on the defaults.

PR #193, opened 2022-09-14
Hotfix/bestuursperiode selector

cecemel:
An attempt at re-wiring the bestuursperiode selector. This PR covers only the code-updates to work with mandatarissen. Let's keep it focused.
- Logic about what periods should be populated is moved outside of the component
- Component only works on periods (an intermediate data structure because no equivalent exists in the data model)
- The calculation of the default period is try to fetch current period, if not the future one, if not the first one
  - A lot of assumptions have been done about what the periods should look like. For non historical data, i.e. data ABB is under control of this should fit the bill. 

I tested it, by checking several select cases, and also making sure that once one eenheid is selected, the mandataris is correctly attached to the bestuursorgaan in the bestuursperiode. Feel free to test these too.

 A next pr will bring this to eredienst mandatarissen.  To keep things focused.
(Currently big copy paste, still reluctant to extract common bits ATM)
Note too: in this PR the selector for eredienst mandatarissen is broken.

### DIFF
--- a/app/components/mandatenbeheer/bestuursperioden-selector.js
+++ b/app/components/mandatenbeheer/bestuursperioden-selector.js
@@ -1,5 +1,4 @@
 import Component from '@glimmer/component';
-import moment from 'moment';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
--- a/app/components/mandatenbeheer/bestuursperioden-selector.js
+++ b/app/components/mandatenbeheer/bestuursperioden-selector.js
@@ -7,7 +7,7 @@ export default class MandatenbeheerBestuursperiodenSelectorComponent extends Com
 
   constructor() {
     super(...arguments);
-    this._options = this.args.options || [];
+    this._options = this.args.options.sortBy('startDate') || [];
   }
 
   @action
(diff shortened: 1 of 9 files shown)


Windvis (approved):

app/components/mandatenbeheer/bestuursperioden-selector.js
```diff
@@ -1,60 +1,17 @@
...
-        return (
-          o.bindingStart.toDateString() ==
-            new Date(this.args.selectedStartDate).toDateString() &&
-          !o.bindingEinde
-        );
-      });
-    } else {
-      return this._options[0];
-    }
+    this._options = this.args.options.sortBy('startDate') || [];
```
Windvis, line 10: Wouldn't it make more sense to do this sorting before passing it into the component?

Also, sorting like this only works because the array prototype extensions are enabled, and that will be deprecated soon. I would use native sorting, or the lodash equivalent https://lodash.com/docs/4.17.15#sortBy.

cecemel: Nitpick 1: bwaaa that's presentation logic seems the place to be for a component no?
nitpick 2: ok native will be

And of course these last minute changes have a potential bug :-)

Windvis: It doesn't have to be changed now (hence the "nitpick"). I just mentioned it as a possible improvement.

Ship it as-is if it is urgent.

Poltergeistz (approved):

PR #194, opened 2022-09-14
Hotfix/bestuursperiode selector eredienst mandatarissen

cecemel:
See #193 but then for eredienstmandatarissen. Perhaps @Poltergeistz it is good to test here.

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/mandatarissen.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandatarissen.js
@@ -42,13 +42,13 @@ export default class EredienstMandatenbeheerMandatarissenController extends Cont
   }
 
   @action
-  selectPeriod(startDate, endDate) {
+  selectPeriod(period) {
     const queryParams = {
       page: 0,
-      startDate: startDate,
+      startDate: period.startDate,
     };
 
-    queryParams['endDate'] = endDate;
+    queryParams['endDate'] = period.endDate;
 
     this.router.transitionTo('eredienst-mandatenbeheer.mandatarissen', {
       queryParams,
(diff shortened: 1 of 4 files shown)


Poltergeistz (commented):
Testing feedback : 

- LGTM for eredienst.
- Looks like when periodes are from the past, we get the one further in the past not the one close to today. (For mandatenbeheer)

That's where I had issues, maybe we can take the difference between the periodes and today. Compare them and the closest to today wins. But I'm afraid that when startingDate overlap (do they ?) we will have multiple periodes. Or maybe there is something more simple ?

cecemel:
> * Looks like when periodes are from the past, we get the one further in the past not the one close to today. (For mandatenbeheer)

You mean e.g. for periods
[2001-2003][2003-2004][2004-2005] the [2001-2003] gets selected?

Poltergeistz:
> > * Looks like when periodes are from the past, we get the one further in the past not the one close to today. (For mandatenbeheer)
> 
> You mean e.g. for periods [2001-2003][2003-2004][2004-2005] the [2001-2003] gets selected? -> if so it should be a minor fix here: https://github.com/lblod/frontend-loket/pull/194/files#diff-a5a2f325a91b7bbcaaaf2c441492ca2b5b7eacd68aa61a830c4399fc08958afdR82 just take the last element in the array?

Yes exactly. But the test case is true for 
```
const currentPeriod = sortedPeriods.find(
        (p) => p.startDate <= today && (today < p.endDate || !p.endDate)
      ); 
```

the p.startDate will always be <= today when the priodes are from the past and with no p.endDate, so it shows the first one in the list in that case [2012 - null].

Windvis:
Yea, I think taking the most recent one if non of the periods match the current date make more sense than picking the first period in time.

cecemel:
> the p.startDate will always be <= today when the priodes are from the past and with no p.endDate, so it shows the first one in the list, in that case [2012 - null] comes first.

Yeah, it assumes no messy data (so the data currently in dev is not representative)
so e.g. [2001 - null] [2002-null][2020-2021][2021-null] is not supported. 
We discussed with business, all historical data we had so far (which provided  messy periods) can be ignored

cecemel:
@Windvis @Poltergeistz will fix in both PR's

PR #198, opened 2022-09-15
[WIP] ExpectedEndDate for worship-mandatee

Poltergeistz:
DL-4194

New field `GeplandeEinddatumAanstelling` is added and is disabled by default.

Concerning the mandate `Bestuurslid (van rechtswege) van het bestuur van de eredienst`, since this mandate never ends, currently the expectedEndDate takes the furthest periodeEndDate. We could set a case where if it's this mandate, we show a field with endDate = N/A, but is this going to work on other environments ?

### DIFF
--- a/app/routes/eredienst-mandatenbeheer/new.js
+++ b/app/routes/eredienst-mandatenbeheer/new.js
@@ -27,6 +27,10 @@ export default class EredienstMandatenbeheerNewRoute extends Route {
 
       let worshipMandatee = this.store.createRecord('worship-mandatee');
       worshipMandatee.isBestuurlijkeAliasVan = person;
+      let bestuursperiodesEndDates = tijdsspecialisaties.map(
+        ({ bindingEinde }) => bindingEinde
+      );
+      worshipMandatee.expectedEndDate = bestuursperiodesEndDates[0];
 
       return {
         worshipMandatee,
--- a/app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
+++ b/app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
@@ -63,6 +63,14 @@
         />
       </div>
     </div>
+    <div class="au-u-margin-bottom-small">
+      <AuLabel for="mandate-expected-end-date">GeplandeEinddatumAanstelling</AuLabel>
+      <AuDatePicker
+        @value={{this.model.expectedEndDate}}
+        @disabled={{true}}
+        @id="mandate-expected-end-date"
+      />
+    </div>
   </div>
 
   <AuHr @size="small" />
--- a/app/templates/eredienst-mandatenbeheer/new.hbs
+++ b/app/templates/eredienst-mandatenbeheer/new.hbs
@@ -86,6 +86,15 @@
         </div>
       </div>
 
+      <div class="au-u-margin-bottom-small">
+        <AuLabel for="mandate-expected-end-date">GeplandeEinddatumAanstelling</AuLabel>
+        <AuDatePicker
+          @value={{@model.worshipMandatee.expectedEndDate}}
+          @disabled={{true}}
+          @id="mandate-expected-end-date"
+        />
+      </div>
+
       <AuAlert @skin="info" @icon="info-circle" @size="small">
         <p>
           Klik op "Mandaat toevoegen" om contactgegevens verder aan te vullen.


app/routes/eredienst-mandatenbeheer/new.js
```diff
@@ -27,6 +27,10 @@ export default class EredienstMandatenbeheerNewRoute extends Route {
 
       let worshipMandatee = this.store.createRecord('worship-mandatee');
       worshipMandatee.isBestuurlijkeAliasVan = person;
+      let bestuursperiodesEndDates = tijdsspecialisaties.map(
```
cecemel, line 30: I don't think this will work. I am not sure the parent route calls the bestuursorganen in tijd, but even then, a mandate can be tied to different bestuursorganen. So only at the moment of setting the mandate to the mandataris, there is enough information to deduce the planned enddate. 

My gut feeling tells me it should be close to the setter/save of the mandataris not in the router of new. @Windvis has better opinions .
For sure, it needs to be calculated every time a mandate is selected.

The call is (tested)
```
https://loket.lblod.info/bestuursorganen?filter[bevat][:uri:]=http://data.lblod.info/id/mandaten/04a0d09a6eb812dbc13e84079e2a82eb
```
Also `sort=-binding-start` may work. not sure

(also either use the mu:uuid of uri, not `http://data.lblod.info/id/mandaten/${mandat.id}`. because URI and UUID don't necessarily match)

Furthermore don't forget to _**not**_ set the date for the `org:role` van 'bestuurslid van rechtswege'

app/routes/eredienst-mandatenbeheer/new.js
```diff
@@ -27,6 +27,10 @@ export default class EredienstMandatenbeheerNewRoute extends Route {
...
+        ({ bindingEinde }) => bindingEinde
+      );
+      worshipMandatee.expectedEndDate = bestuursperiodesEndDates[0];
```
cecemel, line 33: Note. Take into account the fact that sometimes It might be missing (i expect them sometimes they will change their mind)

cecemel (changes requested):

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
```diff
@@ -63,6 +63,25 @@
         />
       </div>
     </div>
+    {{#if this.model.expectedEndDate}}
+    <div class="au-u-margin-bottom-small">
+      <AuLabel for="mandate-expected-end-date">GeplandeEinddatumAanstelling</AuLabel>
+      <AuDatePicker
```
Windvis, line 69: Nitpick: I wouldn't use the datepicker here since it might give the impression that it can be selected in some situations while it won't ever be the case in practice.

I would either use the same input component setup you use for the `N/A` case or simply show the value as "plain text" with something like the `DataList` component that we already use in other places.

app/routes/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -26,6 +26,11 @@ export default class EredienstMandatenbeheerMandatarisEditRoute extends Route {
       'filter[type]': CONTACT_TYPE.PRIMARY,
       include: 'adres,secondary-contact-point',
     });
+
+    this.bestuursorganenInTijd = await this.store.query('bestuursorgaan', {
+      'filter[bevat][:uri:]': mandataris.bekleedt.get('uri'),
```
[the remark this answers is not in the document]
Windvis: Also, I don't think we can fetch the data here since the user can still change the mandate on the edite page. I think we need the same setup as on the `new` page?

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -153,4 +154,20 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       this.editingContact = null;
     }
   }
+
+  setExpectedEndDate(mandaat) {
+    if (
+      mandaat.bestuursfunctie.get('label') ===
+      'Bestuurslid (van rechtswege) van het bestuur van de eredienst'
+    ) {
```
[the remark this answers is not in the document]
Poltergeistz: Thanks for the review :raised_hands: 

I've refactored this function to made it exportable in the `/utils` folder along with `fetchBestuursorganenInTijd` function since the route `new` & `edit` share the same logic.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -153,4 +154,20 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
+      this.model.expectedEndDate = undefined; // Bestuurslid (van rechtswege) is a lifetime mandate
+    } else {
+      let plannedEndDates = this.bestuursorganenInTijd.map(
+        (o) => o.bindingEinde
```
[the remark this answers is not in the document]
Poltergeistz: Yes I agree, I'm changing it.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -153,4 +154,20 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
+      );
+      this.model.expectedEndDate = !plannedEndDates
```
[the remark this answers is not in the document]
Poltergeistz: It was a check as some `binding-einde` dates could be expected to be undefined. The bestuursorganenInTijd array does have values sorted by desc order, btw does it consider undefined to be smaller?

I that case, we could only check if the first index is undefined as it would say that every `binding-einde` dates are undefined ? :thinking:

app/routes/eredienst-mandatenbeheer/mandataris/edit.js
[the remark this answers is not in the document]
Poltergeistz: Ok, I'll avoid `.get`

I've removed the fetch from the route and exported directly in the controller with the util function `fetchBestuursorganenInTijd()` so it can be used in `new` & `edit` controllers.

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
Poltergeistz: I've made the change for the `<AuInput />`

app/templates/eredienst-mandatenbeheer/new.hbs
```diff
@@ -86,6 +86,26 @@
         </div>
       </div>
 
+      {{#if @model.worshipMandatee.expectedEndDate}}
```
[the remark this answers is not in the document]
Poltergeistz: Yes, I think we could do that.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -25,8 +29,13 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
   }
 
   @action
-  setMandaat(mandaat) {
+  async setMandaat(mandaat) {
     this.model.bekleedt = mandaat;
+    setExpectedEndDate(
+      await fetchBestuursorganenInTijd(this.store, mandaat.uri),
```
Windvis, line 35: Instead of fetching the data here and passing it in, I think the `setExpectedEndDate` util could do the fetching itself?

So you would pass in the store, mandatee and mandate records and it could do the rest? That would simplify it even more I think?

app/utils/eredienst-mandatenbeheer.js
```diff
@@ -0,0 +1,26 @@
...
+    'http://data.vlaanderen.be/id/concept/BestuursfunctieCode/5972fccd87f864c4ec06bfbd20b5008b'; // Bestuurslid (van rechtswege) is a lifetime mandate
+  let bestuursfunctie = await mandaat.bestuursfunctie;
+  if (bestuursfunctie.uri === lifetimeMandate) {
+    mandataris.expectedEndDate = undefined;
+  } else {
+    let plannedEndDates = bestuursorganenInTijd.map(
+      (bestuursorgan) => bestuursorgan.bindingEinde
+    );
+    mandataris.expectedEndDate =
+      plannedEndDates[0] === undefined ? undefined : plannedEndDates[0];
```
Windvis, line 24: This would be the same as simply setting it to `plannedEndDates[0]` right?
```suggestion
    mandataris.expectedEndDate = plannedEndDates[0];
```

app/routes/eredienst-mandatenbeheer/new.js
Windvis: @cecemel What should happen if the end date is missing?

app/utils/eredienst-mandatenbeheer.js
```diff
@@ -0,0 +1,26 @@
...
+  const lifetimeMandate =
+    'http://data.vlaanderen.be/id/concept/BestuursfunctieCode/5972fccd87f864c4ec06bfbd20b5008b'; // Bestuurslid (van rechtswege) is a lifetime mandate
+  let bestuursfunctie = await mandaat.bestuursfunctie;
+  if (bestuursfunctie.uri === lifetimeMandate) {
+    mandataris.expectedEndDate = undefined;
+  } else {
+    let plannedEndDates = bestuursorganenInTijd.map(
+      (bestuursorgan) => bestuursorgan.bindingEinde
+    );
+    mandataris.expectedEndDate =
```
Windvis, line 23: Also, I'm not 100% sure if this does what we want. I think the logic was more complex than simply taking the first period in time and using that one's end date? I vaguely remember some sort of discussion about this but I can't find it anymore.

@cecemel could you double check if this is the correct behavior?

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
```diff
@@ -63,6 +63,14 @@
         />
       </div>
     </div>
+    <div class="au-u-margin-bottom-small">
```
Windvis, line 66: Do we also want the if check here to make it match the "new" page? Or will the mandate always be set on the edit page? Is selecting a mandate a requirement on the "new" page ?

Windvis:
Also, the `[WIP]` in the PR title is confusing, I was hesitant to review because of it 😄. GitHub also has a "Draft" PR feature that can be used instead.

Poltergeistz:
> Also, the `[WIP]` in the PR title is confusing, I was hesitant to review because of it smile. GitHub also has a "Draft" PR feature that can be used instead.

Ah sorry, wont use this again :sweat_smile:

app/utils/eredienst-mandatenbeheer.js
Poltergeistz: Yes indeed :man_facepalming:

app/utils/eredienst-mandatenbeheer.js
Poltergeistz: Currently, all bestuursorganen in tijd linked to the selected mandate are sorted by DESC order so the greatest (the one far in the future) is the expectedEndDate. As far as I remember, I understood that is the case we want but it's worth to double check.

This leaves some questions : 
We have 2 bestuursperiodes linked to 2 bestuursorganen in tijd. E g : [2023-2026] - [2026-2029]  
What if we add a 3rd, [2029-2031] ?

1- When a mandate is tied to a bestuursorganen in tijd ? 
- After the creation of a new mandataris (with a mandate selected) on the `new` route ? 
- Or after when we select the bestuursperiode and creating a new mandataris ? 
- Or none of this ?

I have the feeling it will be not linked to any mandates yet, so with bestuursperiode [2029-2031] selected, if we create a new mandataris, and if we select any mandates during the step it will not suggest [2031] but [2026] or [2029] depending of the number of tied bestuursorganen in tijd ?

I wonder if there is a way to wipe the data locally because it's tricky for me to check now :confused:

app/routes/eredienst-mandatenbeheer/new.js
cecemel: @Poltergeistz @Windvis not set it and eventually set 'Niet  beschikbaar' in the template

app/utils/eredienst-mandatenbeheer.js
cecemel: @Windvis yeah i think that is correct, from the context, i deduce the bestuursorganen in tijd are passed linked to the mandate. @Poltergeistz i would at least put this assumption in a comment from the setExpectedEndDate (also about the ordering)

`We have 2 bestuursperiodes linked to 2 bestuursorganen in tijd. E g : [2023-2026] - [2026-2029]
What if we add a 3rd, [2029-2031] ?`
-> This shouldn't happen and if it happens we'll have to do correct his by a migration. Hence, not a worry for the fronted. Good asking thanks
Note if bestuuursperiodes are added, they will (very likely) also be linked to new mandates (although with the same bestuursfuncties)

so I think it answers the rest of your worries?

Windvis: Yup, I assumed it was more complex than this, but I couldn't remember the details. Thanks for clarifying!

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
cecemel: What is the discussed check here? 
Anyway,  mandate  can be set on edit and new ? Not sure if it MUST be set on new.

Windvis: To only show that field once a mandate type has been selected. At the moment it will show `N/A` if the mandate type is empty.

As can be seen here:
https://github.com/lblod/frontend-loket/pull/198/files#diff-c393af8ab46b9e79a388e055c1f31eef2f2c698b65f11d3b274096f392a386e7R89-R98

It was a remark I made here: https://github.com/lblod/frontend-loket/pull/198#discussion_r977400358

cecemel: Yeah I think it is fine right? Also from user's perspective less irrelevant data for a new mandataris?

Windvis:
> Ah sorry, wont use this again 😅
Ah no, it's no problem, I do it as well, but in that case I would remove the `[WIP]` from the title if I want it reviewed 😄. The draft PR system just makes that a bit easier because you can just press a button.

app/utils/eredienst-mandatenbeheer.js
Poltergeistz: Okay thanks for the answers :heavy_check_mark:
I'll document that.

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
Poltergeistz: You mean we can create a new mandataris without any mandate set ? How do we select it again when we leave the edit page if we don't set a mandate at this moment ? (forgive my ignorance)

Windvis: > You mean we can create a new mandataris without any mandate set ? (isn't the field mandaat mandatory ? I mean right now it's not but at some point it should be the case right ?)

Yea, it's a strange thing to do. Usually the client prefers as little validation rules as possible, but in this case I think it makes sense to enforce it.

> How do we select the mandataris again when we leave the edit page if we don't set a mandate at this moment ? (forgive my ignorance)

I'm not sure what you mean with this? Select the mandataris where?

Poltergeistz: > Yea, it's a strange thing to do. Usually the client prefers as little validation rules as possible, but in this case I think it makes sense to enforce it.
>
Okay, check is added.
> I'm not sure what you mean with this? Select the mandataris where?
> 
If we try to select a mandataris without any mandaat from the route `/eredienst-mandatenbeheer/mandatarissen`, he simply wont show in the list of mandataris in the data-table since no mandaat mean no periode attached, so we wont be able to select it again unless we have the route url. (this can happen if the field mandaat is not required as mentioned above)

app/utils/eredienst-mandatenbeheer.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+export async function setExpectedEndDate(store, mandataris, mandaat) {
+  const lifetimeMandate =
+    'http://data.vlaanderen.be/id/concept/BestuursfunctieCode/5972fccd87f864c4ec06bfbd20b5008b'; // Bestuurslid (van rechtswege) is a lifetime mandate
+  let bestuursfunctie = await mandaat.bestuursfunctie;
+
+  let bestuursorganenInTijd = await store.query('bestuursorgaan', {
+    'filter[bevat][:uri:]': mandaat.uri,
+    sort: '-binding-start',
+  });
```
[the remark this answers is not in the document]
Poltergeistz: Good point I'll change this, thanks.

Windvis (approved):

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
cecemel: good point, so at least 1 mandataris selected right?

PR #206, opened 2022-09-26
Prevent to save a `mandataris` without a `mandaat`

Poltergeistz:
This sets a check when creating a new `mandataris`.
The user **must** select a `mandaat` from the selector to create a new `mandataris`.

UX Improvements to be made ?

- `Verplicht` pill for mandaat label.
- Error handling to inform the user (or simply disable the save button for now until a mandaat is selected).

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/new.js
+++ b/app/controllers/eredienst-mandatenbeheer/new.js
@@ -36,6 +36,7 @@ export default class EredienstMandatenbeheerNewController extends Controller {
   setMandaat(mandaat) {
     const { worshipMandatee } = this.model;
     worshipMandatee.bekleedt = mandaat;
+    worshipMandatee.errors.remove('bekleedt');
     setExpectedEndDate(this.store, worshipMandatee, mandaat);
   }
 
@@ -56,7 +57,7 @@ export default class EredienstMandatenbeheerNewController extends Controller {
         worshipMandatee.id
       );
     } else {
-      worshipMandatee.errors.add('mandaat', 'Mandaat is een vereist veld.');
+      worshipMandatee.errors.add('bekleedt', 'Mandaat is een vereist veld.');
     }
   }
 }
--- a/app/controllers/eredienst-mandatenbeheer/new.js
+++ b/app/controllers/eredienst-mandatenbeheer/new.js
@@ -3,6 +3,7 @@ import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import { dropTask } from 'ember-concurrency';
+import { isMandaatSelected } from 'frontend-loket/models/worship-mandatee';
 import { setExpectedEndDate } from 'frontend-loket/utils/eredienst-mandatenbeheer';
 
 export default class EredienstMandatenbeheerNewController extends Controller {
@@ -11,7 +12,6 @@ export default class EredienstMandatenbeheerNewController extends Controller {
 
   queryParams = ['personId'];
   @tracked personId = '';
-  @tracked shouldSelectMandaat = true;
 
   get shouldSelectPerson() {
     return !this.model?.person;
@@ -34,7 +34,6 @@ export default class EredienstMandatenbeheerNewController extends Controller {
 
   @action
   setMandaat(mandaat) {
-    this.shouldSelectMandaat = false;
     const { worshipMandatee } = this.model;
     worshipMandatee.bekleedt = mandaat;
     setExpectedEndDate(this.store, worshipMandatee, mandaat);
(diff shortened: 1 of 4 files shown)


app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -48,11 +50,16 @@ export default class EredienstMandatenbeheerNewController extends Controller {
     event.preventDefault();
 
     let { worshipMandatee } = this.model;
-    yield worshipMandatee.save();
+    if (this.shouldSelectMandaat) {
```
Windvis, line 53: I would use an async util for this instead. That way the state can simply be derived from the mandatee record so we don't need to keep a custom flag here.

```suggestion
    if (shouldSelectMandaat(worshipMandatee)) {
```

The util could then simply check if the mandate relationship is set.

app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -48,11 +50,16 @@ export default class EredienstMandatenbeheerNewController extends Controller {
...
+      return {}; // Room for some error handling
```
Windvis, line 54: I would add an error to the `worshipmandatee` record, similar to how we did it for the contact information.

something like (Pseudo) `worshipMandatee.errors.add('mandate', 'Dit veld is verplicht')`

We can then show this error in the template.

Windvis (changes requested):
I would also do the UX suggestions you already mentioned. It should be fairly simple.

- show the required pill in the label
- show an error message
- show a red border on the select. This would also need a new `@error` argument which would should the correct error state on the select.

app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -50,16 +49,14 @@ export default class EredienstMandatenbeheerNewController extends Controller {
...
+    if (yield isMandaatSelected(worshipMandatee)) {
       yield worshipMandatee.save();
-      this.shouldSelectMandaat = true;
-
       this.router.transitionTo(
         'eredienst-mandatenbeheer.mandataris.edit',
         worshipMandatee.id
       );
+    } else {
+      worshipMandatee.errors.add('mandaat', 'Mandaat is een vereist veld.');
```
Poltergeistz, line 59: Is it better to move this line into the util function ? 
Something like `if (!mandate) worshipMandatee.errors.add('mandaat', 'Mandaat is een vereist veld.');`

app/templates/eredienst-mandatenbeheer/new.hbs
```diff
@@ -59,7 +59,7 @@
     <div class="au-u-2-3@medium au-o-box">
       <div class="au-u-margin-bottom-small">
         {{#let (if @model.worshipMandatee.errors.mandaat true false) as |showError|}}
-        <AuLabel for="mandaat" @error={{showError}}>Mandaat</AuLabel>
+        <AuLabel for="mandaat" @error={{showError}} @required={{true}}>Mandaat</AuLabel>
```
Poltergeistz, line 62: Unrelated to this PR but I've noticed in the FP that the mandaat startDate is also mandatory.
Does it makes sense to handle this now or it's not top priority ?

app/controllers/eredienst-mandatenbeheer/new.js
Windvis: Oh, I think it's ok like this for now? It might make sense if we add more validations, but in that case we should turn it into a "validate" method which would add the errors after which we could check the `isValid` property on the record.

app/templates/eredienst-mandatenbeheer/new.hbs
Windvis: It might make sense. I would double check with @cecemel if it should happen now though.

app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -48,11 +49,14 @@ export default class EredienstMandatenbeheerNewController extends Controller {
...
-      worshipMandatee.id
-    );
+    if (yield isMandaatSelected(worshipMandatee)) {
+      yield worshipMandatee.save();
+      this.router.transitionTo(
+        'eredienst-mandatenbeheer.mandataris.edit',
+        worshipMandatee.id
+      );
+    } else {
+      worshipMandatee.errors.add('mandaat', 'Mandaat is een vereist veld.');
```
Windvis, line 59: The first string should be the name of the attribute or relationship (so that would be 'bekleedt' here). It has the benefit that the error message will be cleared if the value changes. Except... It seems that isn't working for relationships at the moment 😄 (which might be a bug in Ember Data). Maybe it's related to the inheritance, not sure.

```suggestion
      worshipMandatee.errors.add('bekleedt', 'Mandaat is een vereist veld.');
```

I would still clear the error message if the user selects a value after the message is shown. For now I would explicitly [remove](https://api.emberjs.com/ember-data/4.7/classes/Errors/methods/remove?anchor=remove) it in the 'setMandate' action.

app/controllers/eredienst-mandatenbeheer/new.js
Poltergeistz: Yes I think it's ok too. Just checking :slightly_smiling_face:

Windvis (approved):

PR #208, opened 2022-09-28
[Mandatenbeheer] Fixing duplicate bestuursorgaan naam

Poltergeistz:
This reuses the same code from #205

Moving the `getUniqueBestuursorganen()` to mandataris class to reuse it in eredienst-mandatenbeheer.

### DIFF
--- a/app/models/mandataris.js
+++ b/app/models/mandataris.js
@@ -20,3 +20,17 @@ export default class MandatarisModel extends AgentInPosition {
   @belongsTo('mandataris-status-code', { inverse: null }) status;
   @hasMany('contact-punt', { inverse: 'mandatarissen' }) contactPoints;
 }
+
+export async function getUniqueBestuursorganen(mandataris) {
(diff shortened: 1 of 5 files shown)


app/models/mandataris.js
```diff
@@ -20,3 +20,17 @@ export default class MandatarisModel extends AgentInPosition {
   @belongsTo('mandataris-status-code', { inverse: null }) status;
   @hasMany('contact-punt', { inverse: 'mandatarissen' }) contactPoints;
 }
+
+export async function getUniqueBestuursorganen(mandataris) {
```
Windvis, line 24: I would also reuse this in the other module so the util code doesn't get out of sync.

Poltergeistz: Yes, in this PR ?

Windvis: Yea. In theory it could have been part of the work to move it to the model class.

Windvis (approved):
Thanks!

PR #210, opened 2022-09-30
Replace "includes" with manual relationship fetching

Windvis:
This should lower the burden on the database while still preventing the data from "popping in" visually.

### DIFF
--- a/app/routes/public-services/index.js
+++ b/app/routes/public-services/index.js
@@ -31,7 +31,6 @@ export default class PublicServicesIndexRoute extends Route {
     let query = {
       'filter[created-by][:uri:]': this.currentSession.group.uri,
       'page[number]': page,
-      include: 'target-audiences,type,executing-authority-levels,status',
     };
 
     if (search) {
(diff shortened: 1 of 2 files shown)


app/routes/public-services/index.js
```diff
@@ -41,6 +40,21 @@ export default class PublicServicesIndexRoute extends Route {
...
+
+    let publicServices = yield this.store.query('public-service', query);
+
+    let promises = [];
+    publicServices.toArray().forEach((service) => {
+      promises.push(
+        service.hasMany('targetAudiences').load(),
+        service.hasMany('executingAuthorityLevels').load(),
+        service.belongsTo('type').load(),
+        service.belongsTo('status').reload()
```
Windvis, line 52: `reload` is needed here since it's a sync relationship and Ember Data expects it to be loaded with the original record. `load` then doesn't actually do anything, so we use reload to force it.

cecemel: Only reload for status? Cause it would nice to reload all relations once they got updated

Windvis: oh, right. It's probably safer to use reload everywhere then. Thanks.

PR #211, opened 2022-09-30
Update ember-rdfa-editor to v0.63.4

Windvis:
Changelog: https://github.com/lblod/ember-rdfa-editor/blob/master/CHANGELOG.md#0634-2022-09-29

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "moment": "^2.29.1"
   },
   "devDependencies": {
-    "@appuniversum/appuniversum": "^1.2.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,7 +33,6 @@
     "moment": "^2.29.1"
   },
   "devDependencies": {
-    "@appuniversum/appuniversum": "^1.2.0",
```
Windvis, line 36: This was only needed to work around a problem in the 0.59.1 version.

PR #217, opened 2022-10-12
Shouldn't select and save an endDate that is less or equal than the starting date

Poltergeistz:
DL-4375

This prevent to save the record if the ending date is less or equal than the starting date for `eredienst-mandatenbeheer` and `bedienarenbeheer`. Sending an error to the user if it's the case.

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -34,11 +34,14 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
   @action
   handleDateChange(attributeName, isoDate, date) {
     this.model[attributeName] = date;
-    if (this.model.einde <= this.model.start) {
-      this.model.errors.add(
-        'einde',
-        'De einddatum moet na de startdatum liggen'
-      );
+    let { start, einde } = this.model;
+    if (einde instanceof Date && start instanceof Date) {
+      if (einde <= start) {
+        this.model.errors.add(
+          'einde',
+          'De einddatum moet na de startdatum liggen'
+        );
+      }
     }
   }
 
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -41,6 +41,8 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
           'einde',
           'De einddatum moet na de startdatum liggen'
         );
+      } else {
+        this.model.errors.remove('einde');
       }
     }
   }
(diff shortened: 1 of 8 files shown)


app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -34,6 +34,12 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
   @action
   handleDateChange(attributeName, isoDate, date) {
     this.model[attributeName] = date;
+    if (this.model.einde <= this.model.start) {
```
Windvis, line 37: This will throw false positives in some cases. For example if the end date is cleared it will show the warning. I think this check is only needed if both the start and end date are a date object?

 But it's probably better to validate this before saving, similar to the other validations?

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -34,6 +34,12 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
+      this.model.errors.add(
+        'einde',
+        'De einddatum moet groter zijn dan de begindatum'
```
Windvis, line 40: I'm not sure about the error message but I'm not the best writer. Maybe something like this: 'De einddatum moet na de startdatum liggen'

app/templates/worship-ministers-management/minister/edit.hbs
```diff
@@ -66,12 +66,22 @@
         />
       </div>
       <div class="au-o-grid__item au-u-1-2">
-        <AuLabel for="worship-minister-correction-end-date">Eindedatum</AuLabel>
+        {{#let (if @model.minister.errors.agentEndDate true false) as |showError|}}
+        <AuLabel for="worship-minister-correction-end-date" @error={{showError}}>Eindedatum</AuLabel>
```
Windvis, line 70: Whoops, I just noticed a typo. It should be `Einddatum` instead of `Eindedatum`.

Windvis (changes requested):

Poltergeistz:
https://github.com/lblod/frontend-loket/pull/217/commits/907452a0d7a29bd61d00164281ab7eecf1f300ae Should remove the case when the date is empty since it's not mandatory to have an end date filled in the FP

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
Windvis: I just noticed that we need to clear the error in some cases. For example:

- choose a start date
- choose an end date that comes before it (or is the same)
- an error message will be shown
- change the start date to a date that comes before the end date
- the error message isn't cleared since the end date field value wasn't changed

Windvis (approved):

PR #218, opened 2022-10-12
Old tables don't have a version number.

cecemel:
We assume if not provided, this is version 1.0


FangioD:
@cecemel Fix is valid, it checks for version, if not found, we may assume that it is version '1.0'

PR #220, opened 2022-10-13
Hide unavailable modules

Poltergeistz:
DL-4376

- hiding all unavailable modules

- renaming the `eredienst mandatenbeheer` module to just `mandatenbeheer`.

### DIFF
--- a/app/components/loket-module-card.hbs
+++ b/app/components/loket-module-card.hbs
@@ -1,3 +1,4 @@
+{{#if @isAvailable}}
 <AuCard @flex="true" ...attributes as |c|>
   {{#if (has-block "title")}}
     <c.header @badgeSkin="brand" @badgeIcon={{@icon}}>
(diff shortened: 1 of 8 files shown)


app/components/loket-module-card.hbs
```diff
@@ -1,3 +1,4 @@
+{{#if @isAvailable}}
```
Windvis, line 1: I don't think the `@isAvailable` argument is needed anymore?

app/templates/index.hbs
```diff
@@ -67,7 +74,8 @@
           </:link>
         </LoketModuleCard>
       </li>
-      {{#if (is-feature-enabled "worship-minister-management")}}
+      {{/if}}
+      {{#if (or is-feature-enabled "worship-minister-management" this.currentSession.canAccessWorshipMinisterManagement)}}
```
Windvis, line 78: I think this needs to be an `and` right? The module should only be shown if both checks are true.

Also, `is-feature-enabled` is a helper so it needs to be wrapped in parentheses if nested
```suggestion
      {{#if (and (is-feature-enabled "worship-minister-management") this.currentSession.canAccessWorshipMinisterManagement)}}
```

Windvis (changes requested):

app/components/loket-module-card.hbs
Poltergeistz: It was used to display the `AuAlert` component, it's not needed indeed. I'm changing the test too

Windvis (approved):

PR #221, opened 2022-10-14
Eredienst mandatenbeheer / bedienarenbeheer: Person creation prefill data

Poltergeistz:
DL-4384

This prefill seach data from `/new` route in `/new-person` route for the following inputs : 
- first name
- last name
- rrn

The `rrn` input was also modified to be consistent with the `create-persoon` component.

### DIFF
--- a/app/components/shared/persoon/create-persoon.js
+++ b/app/components/shared/persoon/create-persoon.js
@@ -18,10 +18,10 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
   @service store;
 
   @tracked geslacht;
-  @tracked voornaam;
-  @tracked familienaam;
+  @tracked voornaam = this.args.prefilledValues.voornaam;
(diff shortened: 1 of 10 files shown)


Windvis (changes requested):

Poltergeistz:
Thanks for the review. I'm on it :+1:

Poltergeistz:
I've noticed we can pass an empty string, any character or spaces. Not sure if this is what we want in the future.

Windvis:
> I've noticed we can pass an empty string, any character or spaces. Not sure if this is what we want in the future.

Yea, it probably makes sense to trim the whitespace before creating the "entered data" object? I wouldn't bother with stripping out other characters, the user can still edit the data in the input fields afterwards. And it might affect people with non-European names by accident 😄.

Windvis (approved):

PR #222, opened 2022-10-17
`create-persoon` component: rijksregisternummer UX improvements

Poltergeistz:
DL-4385

This autocomplete the following fields when `rrn` is provided : 

- Geboortedatum
- Geslacht

Handling errors cases on save for now.

### DIFF
--- a/app/components/shared/persoon/create-persoon.js
+++ b/app/components/shared/persoon/create-persoon.js
@@ -8,6 +8,7 @@ import {
   getBirthDate,
   isBiologicalFemale,
   isBiologicalMale,
+  isValidRijksregisternummer,
 } from 'frontend-loket/utils/rijksregisternummer';
 
 const maleId = '5ab0e9b8a3b2ca7c5e000028';
@@ -59,21 +60,6 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
 
   get isFemale() {
     return this.geslacht === femaleId;
-  }
-
-  get isValidRijksregisternummer() {
-    let rr = this.rijksregisternummer;
-    if (isBlank(rr)) return false;
-    if (rr.length != 11) {
-      return false;
-    }
-    const preNillies =
-      parseInt(rr.slice(9, 11)) === 97 - (parseInt(rr.slice(0, 9)) % 97);
-    const postNillies =
-      parseInt(rr.slice(9, 11)) ===
-      97 - ((2000000000 + parseInt(rr.slice(0, 9))) % 97);
-
-    return preNillies || postNillies;
   }
 
   get isNationalityFieldRequired() {
@@ -138,7 +124,7 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
       errors.geboorte = 'geboortedatum is een vereist veld.';
     }
 
-    if (!this.isValidRijksregisternummer) {
+    if (!isValidRijksregisternummer(this.rijksregisternummer)) {
       errors.rijksregisternummer = 'rijksregisternummer is niet geldig.';
     }
     if (
--- a/app/components/shared/persoon/create-persoon.js
+++ b/app/components/shared/persoon/create-persoon.js
@@ -175,7 +175,7 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
     this.rijksregisternummer = rijksregisternummer;
     if (isValidRijksregisternummer(this.rijksregisternummer)) {
       this.birthDate = isBirthDateKnown(this.rijksregisternummer)
-        ? new Date(`${getBirthDate(this.rijksregisternummer)} 00:00:00`)
+        ? new Date(getBirthDate(this.rijksregisternummer))
         : this.birthDate;
       if (isGenderKnown(this.rijksregisternummer)) {
         isBiologicalFemale(this.rijksregisternummer)
--- a/app/components/shared/persoon/create-persoon.js
+++ b/app/components/shared/persoon/create-persoon.js
@@ -8,6 +8,8 @@ import {
   getBirthDate,
   isBiologicalFemale,
   isBiologicalMale,
+  isBirthDateKnown,
+  isGenderKnown,
   isValidRijksregisternummer,
 } from 'frontend-loket/utils/rijksregisternummer';
 
@@ -173,12 +175,16 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
   setRijksregisternummer(rijksregisternummer) {
     if (rijksregisternummer) {
       this.rijksregisternummer = rijksregisternummer;
-      this.birthDate = getBirthDate(this.rijksregisternummer);
-      if (isBiologicalFemale(this.rijksregisternummer)) {
-        this.setGender(femaleId);
-      }
-      if (isBiologicalMale(this.rijksregisternummer)) {
-        this.setGender(maleId);
+      if (isValidRijksregisternummer(this.rijksregisternummer)) {
+        this.birthDate = isBirthDateKnown(this.rijksregisternummer)
+          ? getBirthDate(this.rijksregisternummer)
+          : this.birthDate;
+        if (isGenderKnown(this.rijksregisternummer)) {
+          isBiologicalFemale(this.rijksregisternummer) &&
+          !isBiologicalMale(this.rijksregisternummer)
+            ? this.setGender(femaleId)
+            : this.setGender(maleId);
+        }
       }
     } else {
       this.rijksregisternummer = '';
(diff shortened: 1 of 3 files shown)


app/components/shared/persoon/create-persoon.js
```diff
@@ -177,4 +191,20 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
...
+      if (isBiologicalFemale(this.rijksregisternummer)) {
+        this.setGender(femaleId);
+      }
+      if (isBiologicalMale(this.rijksregisternummer)) {
+        this.setGender(maleId);
+      }
+    } else {
+      this.rijksregisternummer = '';
+    }
+  }
```
Windvis, line 209: We should also check if the rijksregisternummer is valid before doing all the birthdate and gender checking. It will throw a lot of exceptions otherwise.

Also there are edge cases where dates of birth and genders aren't known when the RRN is created so we should check for that first before setting the corresponding values. The library has the `isGenderKnown` and `isBirthDateKnown` utils for that.

Windvis (changes requested):

app/components/shared/persoon/create-persoon.js
```diff
@@ -136,6 +141,15 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
...
     }
+    if (
+      (this.isFemale !== isBiologicalFemale(this.rijksregisternummer) ||
+        this.isMale !== isBiologicalMale(this.rijksregisternummer)) &&
+      this.geslacht &&
+      this.rijksregisternummer
+    ) {
+      errors.geslacht =
+        'het geslacht komt niet overeen met uw rijksregisternummer.';
+    }
```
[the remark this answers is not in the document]
Poltergeistz: I was hesitant to put this actually, I can remove it.

app/utils/rijksregisternummer.js
```diff
@@ -0,0 +1,106 @@
...
+    ]; // Eg. ['86', '08', '14'] from '860814'
+    const serial = normalizedNrn.slice(6, 9); // Eg. '413' from '86081441359'
+    const checksum = normalizedNrn.slice(9, 11); // Eg. '59' from '86081441359'
+    return { birthDate, serial, checksum };
+  }
+  if (typeof nrn === 'object') {
+    return nrn;
+  }
+  throw new Error('Could not parse nrn of invalid type');
+}
```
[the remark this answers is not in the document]
Poltergeistz: I don't know, I don't see any license so it's hard to tell. I guess a clear mention of the repo and the owner is needed.

app/components/shared/persoon/create-persoon.js
```diff
@@ -136,6 +141,15 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
     if (!this.isValidRijksregisternummer) {
```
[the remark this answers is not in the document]
Poltergeistz: Indeed.

app/components/shared/persoon/create-persoon.js
Poltergeistz: It seems there is more errors showing up when I'm trying to test the case where there is no gender or no birth date. I tried to generate a rrn but the website does not offer the option to remove either the date or the gender.

app/components/shared/persoon/create-persoon.js
```diff
@@ -177,4 +170,24 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
...
     this.birthDate = date;
   }
+
+  @action
+  setRijksregisternummer(rijksregisternummer) {
+    if (rijksregisternummer) {
+      this.rijksregisternummer = rijksregisternummer;
+      if (isValidRijksregisternummer(this.rijksregisternummer)) {
+        this.birthDate = isBirthDateKnown(this.rijksregisternummer)
+          ? getBirthDate(this.rijksregisternummer)
```
Windvis, line 180: I think this needs to be an actual Date instance since the `loadOrCreateGeboorte` method expects it to be one. I think just wrapping it in a `new Date(getBirthDate(this.rijksregisternummer))` should work.

Alternatively we could do it in the util itself, but I think keeping that logic where it's needed is better. The string version has the advantage that we don't have to mess with timezones if that isn't needed.

app/utils/rijksregisternummer.js
```diff
@@ -0,0 +1,136 @@
...
+
+  THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+const BIS_MONTH_INCREMENT_GENDER_UNKNOWN = 20;
+const BIS_MONTH_INCREMENT_GENDER_KNOWN = 40;
+
+const mod97 = (input) => String(97 - (Number(input) % 97)).padStart(2, '0');
+
+export function isValidRijksregisternummer(nrn) {
+  const { birthDate, serial } = parse(nrn);
```
Windvis, line 15: Nitpick: I would move this below the first checks since it's only needed at the end of the util.

app/utils/rijksregisternummer.js
```diff
@@ -0,0 +1,136 @@
...
+  if (!(typeof nrn === 'string' && nrn.trim().length > 0)) return false;
+  if (nrn.length !== 11) {
+    return false;
+  }
+  if (isBisNumber(nrn)) return false;
```
Windvis, line 20: I'm not sure about this? I think bis numbers are still valid RRNs, no?

app/utils/rijksregisternummer.js
```diff
@@ -0,0 +1,136 @@
...
+  const preNillies =
+    parseInt(nrn.slice(9, 11)) ===
```
Windvis, line 22: The parse util also returns the checksum, so we can reuse that here I think
```suggestion
    parseInt(checksum) ===
```

And then retrieve that from the parse util:
`const { birthDate, serial, checksum } = parse(nrn);`

app/components/shared/persoon/create-persoon.js
```diff
@@ -177,4 +170,24 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
   setDateOfBirth(isoDate, date) {
     this.birthDate = date;
   }
+
+  @action
+  setRijksregisternummer(rijksregisternummer) {
+    if (rijksregisternummer) {
+      this.rijksregisternummer = rijksregisternummer;
```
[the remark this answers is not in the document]
Poltergeistz: Yes `rijksregisternummer` must be a string or we get errors.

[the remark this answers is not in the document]
Windvis: Yes, but it will always be one already? The input event will return an empty string if no data was entered.

app/utils/rijksregisternummer.js
Poltergeistz: Well me too, I'm not familiar with this. I though BIS and RRN are two different things.

app/utils/rijksregisternummer.js
Poltergeistz: Indeed :+1:

app/utils/rijksregisternummer.js
Windvis: I would just drop that check and keep the same behavior as the previous getter. I'm not 100% sure what the BIS numbers are, but it sounds like it's some sort of temporary RRN (but still a valid one).

Poltergeistz: The issue I encounter was when you enter a RRN with month above 20 or 40 (That's actually a BIS number then), it can be valid and could return an error where the date is not ISO formatted (e.g. the month = 14)

Windvis: But then the `isBirthDateKnown` util would return false I think?

Poltergeistz: Now yes but previously it wasn't the case from the original function.

Poltergeistz:
After some testing it looks like the person is created with empty values.

PR #224, opened 2022-10-24
When the RRN is prefilled the date of birth and gender fields need to be prefilled as well

Poltergeistz:
DL-4427

This calls this.setRijksregisternummer method in the constructor to fill the birth date and the gender.

### DIFF
--- a/app/components/shared/persoon/create-persoon.js
+++ b/app/components/shared/persoon/create-persoon.js
@@ -48,11 +48,17 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
     this.minDate = hundredYearsAgo;
     this.maxDate = eighteenYearsAgo;
 
-    this.birthDate = new Date(
-      now.getFullYear() - 21,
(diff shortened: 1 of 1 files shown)


app/components/shared/persoon/create-persoon.js
```diff
@@ -48,11 +48,17 @@ export default class SharedPersoonCreatePersoonComponent extends Component {
...
-    );
+    if (this.args.prefilledValues?.rijksregisternummer) {
+      this.setRijksregisternummer(
+        this.args.prefilledValues?.rijksregisternummer
+      );
+    } else {
+      this.birthDate = new Date(
+        now.getFullYear() - 21,
+        now.getMonth(),
+        now.getDay()
```
Windvis, line 59: I just noticed that it's using the week day instead of the date of the month 😬. Good that we're finally dropping it 😄.

Windvis (approved):

PR #237, opened 2022-11-17
Update to Appuniversum v2

Windvis:
Some dependencies still depend on v1 so we use overrides to force v2 everywhere. Once that is resolved we can remove the override again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,13 +27,17 @@
     "test:ember": "ember test"
   },
   "overrides": {
+    "@appuniversum/ember-appuniversum": "^2.0.0",
     "@embroider/macros": "^1.0.0",
     "@embroider/util": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,13 +27,17 @@
...
   "overrides": {
+    "@appuniversum/ember-appuniversum": "^2.0.0",
     "@embroider/macros": "^1.0.0",
     "@embroider/util": "^1.0.0",
     "ember-cli-babel": "$ember-cli-babel",
     "moment": "^2.29.1"
   },
+  "overridesNotes": {
+    "appuniversum": "ember-rdfa-editor and ember-acmidm-login bring in a v1 version. We can remove the override once that is resolved."
+  },
```
Windvis, line 38: We can't add comments to the json file so this is a hacky way around that.

PR #239, opened 2022-11-23
Integrate the Sentry client

Windvis:
To enable the integration the server should replace the `{{SENTRY_DSN}}` string with the real dsn value.

An environment name can also be added by replacing the `{{SENTRY_ENVIRONMENT}}` string but it defaults to 'production'.


app/services/current-session.js
```diff
@@ -42,9 +43,22 @@ export default class CurrentSessionService extends Service {
...
 
+  setupSentrySession() {
+    setUser({ id: this.user.id, ip_address: null });
+    setContext('session', {
+      account: this.account.id,
+      user: this.user.id,
+      group: this.group.id,
+      groupClassification: this.groupClassification.id,
+      roles: this.roles,
+    });
```
Windvis, line 59: This is some extra data that will be displayed in the error interface. I've only used the ids for now, do we want other data as well?

image

app/utils/sentry.js
```diff
@@ -0,0 +1,18 @@
+import { init as initSentry } from '@sentry/ember';
+import config from 'frontend-loket/config/environment';
+
+export function setupSentry() {
+  if (config.sentry.dsn !== '{{SENTRY_DSN}}') {
+    let sentryEnvironment =
+      config.sentry.environment !== '{{SENTRY_ENVIRONMENT}}'
+        ? config.sentry.environment
+        : 'production';
```
Windvis, line 9: I've defaulted to "production" so the production server wouldn't have to edit this value. Alternatively I could make both values required. Any preference?

Windvis:
@cecemel I've sent you an invite to the test organization I made on GlitchTip so you can see what it looks like if you want. I'll probably just delete it after this is merged so we can start from a "clean slate".

Windvis:
This is what the email notification looks like:

image

Pretty barebones but I assume that's because of potential privacy implications? Not a real issue since the issue details page shows everything we need.

app/services/current-session.js
cecemel: yeah, if we self host i don't see a problem, perhaps aim at the URI instead of id, tad easier to query if needed.

app/utils/sentry.js
cecemel: no it's ok, we just should be able to pass values at Runtime with the EMBER_ prefix

app/services/current-session.js
Windvis: I've updated the code and used URIs where available, but the account and "gebruiker" resources don't send it over the API yet.

PR #240, opened 2022-12-01
[Bedienarenbeheer / Eredienst-Mandatenbeheer] Allow addresses outside of Flanders

Poltergeistz:
DL-4462

At this stage, we can create or update an address manually with the `input-manual-address` component in all eredienst-modules where the `contact-information-table` is used. ~~Meaning we can also update manually selected addresses but it will set the `adresRegisterUri` & `adresRegisterId` to null.~~
When the `adres` model `isValid` it will fill the fullAddress _(volledigAdres)_ property with `combineFullAddress(address)` when there is all address information available and set undefined attributes to null before saving.

There is error handling for the `input-manual-address` component but might be improved.

**Issues :** 
- ~~The `AdressenregisterSelector` component does not behave like it should [a6e295f](https://github.com/lblod/frontend-loket/pull/240/commits/a6e295f3e417b44e46c278ebd32c70c8f25f0d9c).~~

Steps to reproduce : 

- search for an address in the `AdressenregisterSelector` for the first time.
- click on an address.
- once the address is selected, all attributes are set to undefined in the ember-inspector. 

**Update SOLVED :** Ember-inspector issue.

## Current full flow story
1) Users can create either a manual address _(outside or Flanders)_ or a selected address _(in Flanders)_.
- The address can be new or fetched _(only for the selected address, manual addresses can't be fetched yet)_.
2) Users can modify a saved manual address record, if users leave the current step it will rollback unsaved changes.
3) Users can switch from manual address to a selected address and reverse, it will replace the `contact-punt.adres` by the one picked but not delete the record from the DB _(Do we want to delete previous linked addresses since it could be linked to an other person ? Edge case : two person working on the same address)_.
4) User can't modify a selected address, it will always rollback the previous record when `isNew` and create a new one or fetch one, this behavior prevents duplicate addresses or mistakes with original addresses from Flanders.

## Component change
- newAdres variable from the `row.js` file is now a tracked property, so we can rollback the previous new record before creating a new one. This prevents the new `adres` record to be stacked each time users selects different addresses in the selector. d9c859c


Windvis:
The issue you are seeing could also be this one: https://github.com/emberjs/ember-inspector/issues/1898

Poltergeistz:
> The issue you are seeing could also be this one: [emberjs/ember-inspector#1898](https://github.com/emberjs/ember-inspector/issues/1898)
> 
> It's supposedly already fixed, but the fixes were done in Ember Data and Ember itself, so we probably need to update.

Did not think about this issue, this looks similar. But when I log for the created record I should access the value of this `newAdres.straatnaam` ?

```
newAdres = this.store.createRecord('adres', adres);
console.log('newAdres', newAdres.straatnaam);
```

I have undefined as a result it looks like the record is created without the `adres` payload even if it got data in it. 

I tried to save a minister and indeed the address attributes fields are just undefined. Might be an async/await issue ?

app/models/adres.js
```diff
@@ -11,3 +11,70 @@ export default class AdresModel extends Model {
...
+    return null;
+  }
+}
+
+export async function isValidAdres(address) {
+  const requiredFields = [
+    'straatnaam',
+    'huisnummer',
+    'postcode',
+    'gemeentenaam',
```
Windvis, line 40: I would map these to either the label values or simply use a generic `Dit veld is verplicht` message instead.

At the moment it looks a bit strange since some labels don't match the property names and they are lowercased. It also only works because it uses Dutch words as properties instead of English ones.

app/models/adres.js
```diff
@@ -11,3 +11,70 @@ export default class AdresModel extends Model {
...
+    busnummer: null,
+    huisnummer: null,
+    straatnaam: null,
+    postcode: null,
+    gemeentenaam: null,
+    land: null,
+    adresRegisterId: null,
+    adresRegisterUri: null,
+    volledigAdres: null,
+  });
```
Windvis, line 65: I wouldn't use `setProperties` here. That only works because the Model class still extends classic EmberObject somewhere in the chain, but that's being phased out. I would simply set each property explicitly in the vanilla js way. 
```js
address.busnummer = null;
address.huisnummer = null;
// ...
```

Alternatively I think [`attributes`](https://api.emberjs.com/ember-data/4.7/classes/Model/properties/attributes?anchor=attributes) can be used to loop over all attributes and set it to `null` that way.

app/models/adres.js
```diff
@@ -11,3 +11,70 @@ export default class AdresModel extends Model {
...
+    busnummer: null,
+    huisnummer: address.huisnummer,
+    straatnaam: address.straatnaam,
+    postcode: address.postcode,
+    gemeentenaam: address.gemeentenaam,
+    land: null,
+    adresRegisterId: null,
+    adresRegisterUri: null,
+    volledigAdres: address.volledigAdres,
+  });
```
Windvis, line 79: This feels very strange to me. It seems we simply want to nullify _some_ attributes, but not all? What's the reasoning behind that? In any case, only nullifying the attributes that need it seems better to me. I would then also give the function a more descriptive name since "updateAddressAttributes" is very generic and doesn't describe what it does.

app/utils/empty-string-to-null.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Sets empty strings of given changeset objects to null to avoid saving them in the database
+ *
+ * @param {Object} changeset Ember changeset
+ * @returns Ember changeset
+ */
+export function setEmptyStringsToNull(changeset) {
```
Windvis, line 7: We're not using changesets so I assume this was copied from OP?

Also, I think the intention is to not serialize the empty strings so there aren't any "empty" triples in the database. If so I think this logic is better suited in the Ember Data serializer.

app/utils/empty-string-to-null.js
```diff
@@ -0,0 +1,17 @@
...
+  let properties = Object.keys(changeset.toJSON());
```
Windvis, line 8: `toJSON` is deprecated and it will be removed in v4: https://deprecations.emberjs.com/ember-data/v3.x#toc_record-toJSON

Since I'm trying hard to get the app ready for v4, I wouldn't introduce new deprecated code 😄.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -202,4 +240,67 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       this.editingContact = null;
     }
   }
+
+  async handleInputToggle() {
```
Windvis, line 244: (I'll try to review this method tomorrow, but I wanted to post the current remarks already.)

app/components/contact-information-table/input-manual-address.hbs
```diff
@@ -0,0 +1,107 @@
...
+      </AuButton>
+    </div>
+</td>
+<td>
+  <div class="au-o-block au-u-margin-bottom">
+  <AuLabel
+  >
+    Busnummer
+  </AuLabel>
+  <AuInput @disabled={{true}}/>
```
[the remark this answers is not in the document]
Poltergeistz: I can enable it, I can't remember why I did this in the first place maybe because of the dropdown in the design.

app/models/adres.js
Poltergeistz: Yes I agree, I can go with the generic one.

app/models/adres.js
Poltergeistz: I think we wont need this function if I replace it with 
```
address.adresRegisterId = null;
address.adresRegisterUri = null;
```

The logic was : 

When we're editing addresses that are selected, it maps the fields in the `manual-input-address` component so the uri and id needs to be removed only if the address is been edited otherwise it will result in duplicates. Right now, if we enter manual mode it will set those attributes to `null` but I think we want a trigger (Like an actual change of any property to nullify the props ?).

app/utils/empty-string-to-null.js
Poltergeistz: Ah sorry :confused: I will change this

app/utils/empty-string-to-null.js
Poltergeistz: Yes, okay.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
Poltergeistz: Thanks for the review :smiley:

app/models/adres.js
Poltergeistz: Agreed, plus I would delete this function as it's not called in any file now.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -19,6 +25,11 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
 
   @tracked selectedContact;
   @tracked editingContact;
+  /* is it better to store isManualAddress from the transition.data.isManualAddress
+   in the route ? right now it keeps the current value when we leave the page
+   don't know if we want this behavior.
+  */
+  @tracked isManualAddress = false;
```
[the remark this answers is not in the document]
Poltergeistz: In `addNewContact` since the record is new it wont have any `adresRegisterUri`. So it will trigger manual input mode, is it what we want by default instead of selecting an address from the selector ?

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
Windvis, line 244: I'm not sure about this whole method logic to be honest. Is this how it was implemented in OP? Was this behavior discussed somewhere?

I'll have to look into it some more, but I think it can be simplified on first sight.

Windvis (changes requested):

app/components/contact-information-table/input-manual-address.hbs
```diff
@@ -0,0 +1,107 @@
...
+    <div class="au-o-block">
+      <AuButton @skin="link" {{on "click" @toggleAddressSelector}}>
+        Vind adres in lijst
+      </AuButton>
+    </div>
+</td>
+<td>
+  <div class="au-o-block au-u-margin-bottom">
+  <AuLabel
+  >
```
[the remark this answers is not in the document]
Poltergeistz: I use Prettier but sometimes it does not format on save, that might explain why I have funky indentation so I have to use `ctrl` + `shift` + `P`

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -202,4 +240,67 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
+        (adres) => adres.hasDirtyAttributes
+      ); // Case we have more than one adres we must pick the one with no data
+      if (await currentAddress.id) {
+        console.log('current address manual update');
+        newAddress = fetchAddresses.firstObject;
+      } else {
+        console.log('Selected Address case', filteredAddress);
+        filteredAddress[0].rollbackAttributes(); // Cleaning the data
+      }
+    }
```
Windvis, line 304: If I understand it correctly we want to set the address to the original address that was selected before entering the manual input mode? If so, I think we already stored it on the `originalContactAdres` property if it was persisted before. If we also want to restore the adres that was selected but wasn't saved before switching to the manual input mode then I would store that on a controller property as well.

If it's for cleanup purposes I think we can simply remove the `currentAddress` record? I'm not sure what all this `filtered` logic is about 😄.

Windvis (changes requested):

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -58,6 +69,16 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
   }
 
+  @action
+  async toggleInputMode() {
+    this.isManualAddress = !this.isManualAddress;
+    if (this.isManualAddress) {
+      // Updating a relationship value doesn't seem to clear the corresponding error messages, so we do it manually
+      this.editingContact.errors.remove('adres');
+    }
+    await this.handleInputToggle();
```
Windvis, line 79: nitpick, I don't really see the added value of splitting this of in a separate method. If anything I think it makes things harder to follow since the names are similar (but generic). I would move the code of that method back here.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -111,7 +140,16 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
         }
 
         if (adres?.isNew) {
```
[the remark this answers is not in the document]
Poltergeistz: Yes indeed just catched this too.

app/utils/empty-string-to-null.js
Poltergeistz: I have some issues setting up the serializer. I created a file with the logic we need inside `/serializers` but I don't know how to actually apply the logic to the adres model.

Windvis: If you want to limit the logic to the adres model the file should be at `app/serializers/adres.js`. Ember Data will try to match the serializer name to the model name by default, and fall back to the application serializer.

Poltergeistz: Ok that what I was doing but maybe my logic inside the `adres.js` is wrong. 

```
  serializeAttribute(snapshot, json, attributeName) {
     if (
       !(typeof attributeName === 'string' && attributeName.trim().length > 0)
     ) {
       attributeName = null;
     }
   }
```

Windvis: `attributeName` is the actual name of the attribute, so for example `straatNaam`. Not the value stored for that attribute.

You can receive the value from the [snapshot](https://api.emberjs.com/ember-data/4.7/classes/Snapshot/methods/attr?anchor=attr). So `let value = snapshot.attr(key);`

But I'm not sure if you can modify the value in that method so I assume we must tweak the json directly:

```js
  serializeAttribute(snapshot, json, attributeName) {
    super.serializeAttribute(...arguments);

    let attributeValue = snapshot.attr(attributeName);
   
     if (
       typeof attributeValue === 'string' && attributeValue.trim().length === 0)
     ) {
       json.attributes[attributeName] = null;
     }
   }
```

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -202,4 +240,67 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       this.editingContact = null;
     }
   }
+
+  async handleInputToggle() {
+    let newAddress;
+    let currentAddress = await this.editingContact.adres;
+    let fetchAddresses = this.store.peekAll('adres');
```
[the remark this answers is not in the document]
Poltergeistz: I used this to retrieve the dirty address with the filteredAddress but I think I can simply use currentAddress instead ?

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
Poltergeistz: Sure, I can set it back in the `toggleInputMode()` if it's easier to maintain.

cecemel (changes requested):
Hey, a bit of a stretch to review the code, but I played around.
It gets to do weird things when (errors) when:
- I switch from a previously manual address to a registry addres.
- After this error (i assume broken state) no way to save an address from the registry.

Minor thing, perhaps as a user, if I toggle from fully completed manual address, to an address from the registry, I would not expect the select box to be filled in.

Also note: It seems depending on app-universum 1.5, but now master is on 2.0, perhaps best checking if everything works with that.

Windvis (changes requested):

app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -88,15 +90,54 @@ export default class EredienstMandatenbeheerNewController extends Controller {
...
     this.model.worshipMandatee.errors.remove('contacts');
 
     let primaryContactPoint = createPrimaryContactPoint(this.store);
     let secondaryContactPoint = createSecondaryContactPoint(this.store);
 
     primaryContactPoint.secondaryContactPoint = secondaryContactPoint;
     this.editingContact = primaryContactPoint;
+    if (this.isManualAddress) {
+      this.editingContact.adres = this.store.createRecord('adres');
+    }
```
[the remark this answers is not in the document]
Windvis: I pushed a commit that removes this.

app/controllers/eredienst-mandatenbeheer/new.js
```diff
@@ -111,6 +152,11 @@ export default class EredienstMandatenbeheerNewController extends Controller {
 
     this.originalContactAdres = await contactPoint.adres;
     this.editingContact = contactPoint;
+    if (!(await this.originalContactAdres.adresRegisterUri)) {
```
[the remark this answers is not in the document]
Windvis: Same here.

app/utils/empty-string-to-null.js
Windvis: I'm confused, I think this code is missing from the PR?

Windvis:
> Minor thing, perhaps as a user, if I toggle from fully completed manual address, to an address from the registry, I would not expect the select box to be filled in.

I think this still is the case at the moment? I think clearing the previous address when switching "modes" will simplify the code a lot and that would also cover that remark.

PR #247, opened 2023-01-04
[Worship-Modules] Add link to Databank Erediensten

Poltergeistz:
DL-4240

Adding the right access link for databank Erediensten from the right environment.
Adding the module `databank Erediensten` to be only available for users who does have access to Toezicht.

### DIFF
--- a/app/components/shared/compact-menu.hbs
+++ b/app/components/shared/compact-menu.hbs
@@ -60,7 +60,7 @@
       Producten- en dienstencatalogus
     </AuLink>
   {{/if}}
-  {{#if this.currentSession.canAccessToezicht}}
+  {{#if this.currentSession.canAccessWorshipDecisionsDb}}
     <AuLinkExternal
       @icon="link-external"
       @iconAlignment="left"
--- a/app/components/shared/compact-menu.hbs
+++ b/app/components/shared/compact-menu.hbs
@@ -64,7 +64,7 @@
     <AuLinkExternal
       @icon="link-external"
       @iconAlignment="left"
-      href={{this.currentSession.databankEredienstenUrl}}
+      href={{(worship-decisions-database-url)}}
       role="menuitem"
     >
       Databank Erediensten
--- a/app/components/shared/compact-menu.hbs
+++ b/app/components/shared/compact-menu.hbs
@@ -60,4 +60,14 @@
       Producten- en dienstencatalogus
     </AuLink>
   {{/if}}
+  {{#if this.currentSession.canAccessToezicht}}
+    <AuLinkExternal
+      @icon="link-external"
+      @iconAlignment="left"
+      href={{this.currentSession.databankEredienstenUrl}}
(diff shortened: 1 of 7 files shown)


app/templates/login.hbs
```diff
@@ -214,6 +214,28 @@
               </div>
             {{/if}}
 
+            {{#if this.currentSession.canAccessToezicht}}
```
Windvis, line 217: I believe this will always hide the block since the user isn't logged in on this page. I would either drop the code or always show it.

config/environment.js
```diff
@@ -65,6 +65,9 @@ module.exports = function (environment) {
     ENV.features['eredienst-mandatenbeheer'] = true;
     ENV.features['public-services'] = true;
     ENV.features['worship-minister-management'] = true;
+
+    ENV.databankEredienstenUrl =
```
Windvis, line 69: This will only set the link for local development builds but needs to be configurable from the hosting server, so it needs to be something like this instead:

```js
ENV.worshipDecisionsUrl = '{{WORSHIP_DECISIONS_URL}};
```

The backend can then configure this by setting the `EMBER_WORSHIP_DECISIONS_URL` env variable to a proper value.

Windvis: I would drop the hardcoded dev and prod links from this file as well.

app/components/shared/compact-menu.hbs
```diff
@@ -60,4 +60,14 @@
       Producten- en dienstencatalogus
     </AuLink>
   {{/if}}
+  {{#if this.currentSession.canAccessToezicht}}
```
Windvis, line 63: instead of using the `canAccessToezicht` directly, it could be nice to create a wrapper getter in the session service that also checks the value of the url.

```suggestion
  {{#if this.currentSession.canAccessWorshipDecisionsDb}}
```

```js
get canAccessWorshipDecisionsDb() {
  return this.canAccessToezicht && !config.worshipDecisionsUrl.startsWith('{{');
}
```
(an actual url check is also an option but might be overkill)

That way the template code makes more sense (using `canAccessToezicht` here looks like a copy pasta mistake if you don't know the ticket history) and it won't be shown if the url isn't configured properly.

config/environment.js
Windvis, line 69: I would use English as much as possible. I'm trying hard to Englishify the codebase so it's easier to understand by non-dutch speaking contributors 😄.

Windvis (changes requested):

config/environment.js
Poltergeistz: Understood :slightly_smiling_face:

app/templates/login.hbs
Poltergeistz: Yes indeed, I think we actually don't need it there since as I understood we have to only show the link to a group of users.

Windvis (approved):

PR #248, opened 2023-01-05
Update to Ember 4.8 LTS

Windvis:
This updates the app from 3.28 to 4.8 LTS.

**Open issues:**
None, when clicking through the app and comparing it to QA. But I might have missed some flows and there are no tests, so it's always a gamble.

<details>
  <summary><b>Resolved issues</b></summary>

- some tests need to be updated so they use `this.something` or `@something`. This was deprecated in Ember 3 but the linters didn't check test templates (yet) so it went unnoticed. (See the test job in CI)
- Some navigation links trigger a `[Bug] Missing action` exception. For example, [this link](https://github.com/lblod/frontend-loket/blob/d152c4850937b63994780c1f95c105cfb606771b/app/templates/supervision/submissions/index.hbs#L55) triggers it. It seems to be related to the link being rendered in a AuDataTable and [this action modifier](https://github.com/appuniversum/ember-appuniversum/blob/99f3ceadf9283f5a2f46f49c82327afcf1f1dec9/addon/components/au-data-table/content-body.hbs#L6). Removing that modifier removes the exception so maybe changing that to the `{{on}}` modifier is a quick fix. Weirdly enough, the exception also dissapears when the specific loading template is removed 🤔. "**Fixed" in Appuniversum**
- Ember Data v4.8 in combination with Ember Concurrency tasks triggers `too much recursion` exceptions. Pinning to 4.7 resolves that for now, but that is only a temporary solution. **Resolved on the latest 4.8 patch**
- The `batch-edit` branch isn't Ember v4 ready yet. We need to convert that into a feature flag first to make maintenance easier. #288 
- `record.destroyRecord()` now also runs `unloadRecord` which causes a rerender (with empty data) in some places where we don't expect it. A workaround I've found is to use deleteRecord + save + unloadRecord manually. The unloadRecord part can then be done after the redirect so getters aren't re-executed with an "empty" record.
- Some Ember Data code that worked previously doesn't anymore. 1 such example I found is [this line](https://github.com/lblod/frontend-loket/blob/d152c4850937b63994780c1f95c105cfb606771b/app/models/subsidy-application-flow.js#L11). Accessing an async relationship like was never a good idea, but it seemed to have worked so it went unnoticed. The solution would be to handle the async in a better place and pass the sorted version around where needed. Most likely there are more cases like this. This specific case has been resolved.
- There are also a lot of deprecation warnings triggered by Ember Data, but those can be resolved in separate PRs.
</details>


.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
+name: CI
```
Windvis, line 1: This file needs to be removed, we don't use Github Actions.

Windvis:
I created a new `ember-v4-update` branch which will be used as the long-lived branch for all the update work. We'll only merge that back into dev when everything works as expected.

PR #254, opened 2023-01-11
[Eredienst-mandatenbeheer] View only mode

Poltergeistz:
In cases where the data is provided by the vendor, the bestuurseenheid (administrative-unit) using this vendor should have a view only mode in loket.

- Adding a new read-only page `details` that shows worship-mandatee's data. a49371dc1c72e40f35f61c772fb2f75b6520e61a
- Adding a new attribute `hasEditRight` in the bestuurseenheid model with `defaultValue: true`. a7185660fe5e58c3c2ddbc6be0ad735d67f0aff2
- Redirecting from `eredienst-mandatenbeheer.mandataris.edit`when the bestuurseenheid has no edit rights. 3e9d2b5f58e54f43dbe47edaf33b8ab171ff4a29
- Overview page `eredienst-mandatenbeheer.mandatarissen` is now checking the `hasEditRight` flag. The `Voeg mandataris toe` button is now hidden & the `bewerk` link in the table is now changed to `bekijk`, the link is now going to the details page when the bestuurseenheid has no edit right. b6fe2b536153d93ec023b409f36c1c04151ce72a

Works with https://github.com/lblod/app-digitaal-loket/pull/323


app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
```diff
@@ -0,0 +1,103 @@
...
+
+  {{#if this.contactList}}
+    <div class="au-o-box">
+      <h3 class="au-u-h3 au-u-padding-bottom">
+        Contactgegevens
+      </h3>
+      {{#each this.contactList as |contact|}}
+        <div class="au-o-grid au-o-grid--tiny au-u-margin-bottom-small">
+          <div class="au-o-grid__item au-u-1-2">
+            <AuLabel>Adres</AuLabel>
```
Windvis, line 76: Nitpick: the semantically correct html structure for this is a [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).

[We already do something like this](https://github.com/lblod/frontend-loket/blob/a417340ecf963d3086039a3d449d043c58556e03/app/components/rdf-form-fields/climate-subsidy-costs-table.hbs#L1-L18) so you can use the same setup I think.

app/models/bestuurseenheid.js
```diff
@@ -7,6 +7,7 @@ export default class Bestuurseenheid extends Model {
   @attr mailAdres;
   @attr wilMailOntvangen;
   @attr isTrialUser;
+  @attr({ defaultValue: true }) hasEditRight;
```
Windvis, line 10: Do we know more about how this will actually work in practise?

Windvis (changes requested):

app/models/bestuurseenheid.js
Poltergeistz: No not yet, I just did this setup for testing.

app/routes/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -7,9 +7,15 @@ import {
...
   @service store;
+  @service router;
 
   async beforeModel() {
     const mandatenbeheer = this.modelFor('eredienst-mandatenbeheer');
+    let bestuurseenheid = mandatenbeheer.bestuurseenheid;
+
+    if (!bestuurseenheid.hasEditRight) {
+      this.router.transitionTo('eredienst-mandatenbeheer.mandataris.details');
+    }
```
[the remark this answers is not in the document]
Poltergeistz: I think I can replace this bestuurseenheid variable above with this.currentSession.group.hasEditRight. Also by overview page you mean `eredienst-mandatenbeheer.mandatarissen` ?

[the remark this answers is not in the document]
Windvis: Yes 👍

app/routes/eredienst-mandatenbeheer/mandataris/details.js
```diff
@@ -0,0 +1,39 @@
...
+    const persoon = await mandataris.isBestuurlijkeAliasVan;
+
+    let positionContacts = await mandataris.contacts;
+    this.selectedContact = findPrimaryContactPoint(positionContacts);
+
+    this.contacts = await this.store.query('contact-punt', {
+      'filter[agents-in-position][is-bestuurlijke-alias-van][id]': persoon.id,
+      'filter[type]': CONTACT_TYPE.PRIMARY,
+      include: 'adres,secondary-contact-point',
+    });
```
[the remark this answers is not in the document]
Poltergeistz: Yes I realized I can get the contacts relationship from the worship-mandatee.

app/routes/eredienst-mandatenbeheer/mandataris/edit.js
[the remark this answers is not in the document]
Poltergeistz: Also did the same for the `/new-person` route

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
```diff
@@ -0,0 +1,104 @@
...
+  </dl>
+
+  <AuHr @size="small" />
+
+  {{#if @model.contacts}}
+    <div class="au-o-box">
+      <AuHeading @skin="2" class="au-u-h3 au-u-padding-bottom">
+        Contactgegevens
+      </AuHeading>
+      {{#each @model.contacts as |contact|}}
```
Windvis, line 77: I don't think this is safe to do. AFAIK the `contacts` relationships contains both the primary and secondary contactpoint linked to this position.

In theory this should print 2 blocks of information. This can be verified by adding a new contactpoint to the position with both primary and secondary information. In that case 2 blocks will be shown, 1 shows the primary information, and the other shows the secondary phone number as the primary phone number.

Something is wrong in our edit logic though, because adding a secondary phone number to an already existing contact doesn't trigger this. I think we forget to update the contacts relationship somewhere in the save flow.

In any case, I would fetch the primary contact point in the route (similar to the edit page) and display that here.

image

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
```diff
@@ -0,0 +1,104 @@
...
+  </Group>
+
+</AuToolbar>
+
+<AuBodyContainer @scroll={{true}}>
+  {{#if this.bestuurseenheid.hasEditRight}}
+    <AuAlert @icon="info-circle" @skin="info" @size="small">Deze mandatarisgegevens zijn alleen leesbaar. Om ze te bewerken klikt u op <AuLink @route="eredienst-mandatenbeheer.mandataris.edit">bewerken</AuLink>.</AuAlert>
+  {{else}}
+    <AuAlert @icon="alert-triangle" @skin="warning" @size="small">Deze mandatarisgegevens zijn alleen leesbaar. Je kunt ze niet bewerken.</AuAlert>
+  {{/if}}
```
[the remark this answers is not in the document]
Poltergeistz: Okay so I remove the first one. About details page, are we keeping it accessible (I assumed this since if we can edit we could also read but that's just some guessing about rules) ?

Thanks :+1:

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
Poltergeistz: So revert to what was done before ?

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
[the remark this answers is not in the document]
Windvis: Yea I would keep it accessible. Other apps that have both a details + edit page usually link to the details page from the overview page table. The details page would then have a "edit button" that links to the edit page. So fairly similar to the alert here, but more intuitive imo. I'm not sure if that's what we want here, but it might just be in the future.

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
Windvis: I'm not sure what was done. But in the edit page we find the ["primaryContactPoint"](https://github.com/lblod/frontend-loket/blob/59a3822be8b0923f3c16e0642dc6dd70be1f1564/app/routes/eredienst-mandatenbeheer/mandataris/edit.js#L21) in the relationship. We should do that here as well so we can display that specific contact instead of looping over the `contacts` relationship. I think that should do what we want.

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
```diff
@@ -0,0 +1,104 @@
...
+    </div>
+  </Group>
+
+</AuToolbar>
+
+<AuBodyContainer @scroll={{true}}>
+  {{#if this.bestuurseenheid.hasEditRight}}
+    <AuAlert @icon="info-circle" @skin="info" @size="small">Deze mandatarisgegevens zijn alleen leesbaar. Om ze te bewerken klikt u op <AuLink @route="eredienst-mandatenbeheer.mandataris.edit">bewerken</AuLink>.</AuAlert>
+  {{else}}
+    <AuAlert @icon="alert-triangle" @skin="warning" @size="small">Deze mandatarisgegevens zijn alleen leesbaar. Je kunt ze niet bewerken.</AuAlert>
```
Windvis, line 25: Suggested copy from the internal chat:

```suggestion
    <AuAlert @icon="alert-triangle" @skin="warning" @size="small">Deze mandatarisgegevens zijn alleen leesbaar. Je kunt ze niet bewerken want ze worden beheerd door uw softwareleverancier.</AuAlert>
```

Poltergeistz: Thanks :slightly_smiling_face:

app/templates/eredienst-mandatenbeheer/mandataris/details.hbs
```diff
@@ -0,0 +1,98 @@
...
+          <div class="au-o-grid__item au-u-1-2">
+            <dt class="au-c-label">Secondair telefoonnummer</dt>
+            <dd>
+              <ValueWithPlaceholder
+                @value={{this.contact.secondaryContactPoint.telefoon}}
+              />
+            </dd>
+          </div>
+        </dl>
+        <AuHr @size="small" />
```
Windvis, line 95: I would remove this line now.

```suggestion
```

Windvis (commented):
1 small remark, but other than that it looks good to me 👍. The flags work as expected.

Windvis (approved):

PR #256, opened 2023-01-11
Add a "Conceptual public service" details page

Windvis:
~~This depends on this PR to render the forms: https://github.com/lblod/lpdc-management-service/pull/21~~ This is now included in the development branch of the backend so just running that should be enough.

The only way to get to the page for now is by manually typing it in the url (or do some inspector trickery, but typing will be easier). You can find the uuid of a concept by going to the "add new product" page and looking in the data tab of the inspector.

### DIFF
--- a/app/components/public-services/concept-details-page.hbs
+++ b/app/components/public-services/concept-details-page.hbs
@@ -1,0 +1,36 @@
+<AuBodyContainer @scroll={{true}}>
+  {{#if this.loadForm.isRunning}}
+    <div class="au-o-box">
+      <p class="au-u-h4 au-u-medium" aria-hidden="true">
+        Formulier aan het laden
+      </p>
+      <AuLoader @padding="small" />
+    </div>
+  {{else if this.loadForm.last.isError}}
+    <div class="au-o-box">
+      <AuAlert
+        @title="Onverwachte fout"
+        @skin="error"
+        @icon="circle-x"
+        @size="small"
+        class="au-u-max-width-small"
+      >
+        <p>
+          Er ging iets fout bij het laden van het formulier, gelieve de helpdesk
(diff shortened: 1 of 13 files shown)


app/components/public-services/concept-details-page.hbs
```diff
@@ -0,0 +1,36 @@
+<AuBodyContainer @scroll={{true}}>
```
Windvis, line 1: This is a simplified copy of the `DetailsPage` component. It might be better to add an `@isConcept` argument to that component and do some things conditionally in there since they are very similar.

app/components/public-services/concept-details-page.js
```diff
@@ -0,0 +1,58 @@
...
+      FORM('Form'),
+      FORM_GRAPHS.formGraph
+    );
+
+    this.form = form;
+    this.formStore = formStore;
+  }
+}
+
+async function fetchFormGraphs(serviceId, formId) {
```
Windvis, line 55: If we keep the 2 separate components we can probably extract some common logic.

app/components/public-services/concept-details-page.hbs
Windvis: We'll keep it as-is for now in case things diverge in the future.

app/routes/public-services/concept-details.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default class PublicServicesConceptDetailsRoute extends Route {
+  @service store;
+
+  async model({ conceptId }) {
+    let concept = await this.store.findRecord(
+      'conceptual-public-service',
+      conceptId,
+      {
+        reload: true,
```
Windvis, line 12: Not needed here.

PR #257, opened 2023-01-11
Feature/update acm and add external apps

cecemel:
Two things covered in the PR:
- update to the new acm context for databankErediensten
-  added access to new module: Eredienst organisaties (same pattern as databankErediensten)

please feel free to correct immediately when it burns

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -171,15 +171,35 @@
               van de eredienst, gemeente- en provincieoverheid en representatief
               orgaan.</:description>
             <:link>
-              <AuLinkExternal
-                @skin="button"
-                href={{(worship-decisions-database-url)}}
-              >
-                Ga naar de databank erediensten (externe link)
-              </AuLinkExternal>
+            <AuLinkExternal
+              @skin="button"
+              href={{(worship-decisions-database-url)}}
+            >
+              Ga naar de databank erediensten (externe link)
+            </AuLinkExternal>
             </:link>
           </LoketModuleCard>
         </li>
+      {{/if}}
+      {{#if this.currentSession.canAccessWorshipOrganisationsDb}}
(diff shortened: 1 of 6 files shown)


app/templates/index.hbs
```diff
@@ -171,16 +171,36 @@
...
       {{/if}}
+      {{#if this.currentSession.canAccessWorshipOrganisationsDb}}
+       <li class="au-o-grid__item au-u-1-2@medium au-u-1-3@large">
+           <LoketModuleCard
+             @icon="link-external"
+             @userManualLink="https://abb-vlaanderen.gitbook.io/handleiding-loket/erediensten/eredienst-organisaties"
+           >
+             <:title>Eredienst Organisaties</:title>
+             <:description>Consulteer hier de mandatarissen en bedienaren
+             van de besturen eredienst en centrale besturen. </:description>
```
Windvis, line 192: Current version sounds a bit weird, not sure if this is better:
```suggestion
             van de eredienst besturen en centrale besturen. </:description>
```

app/templates/index.hbs
```diff
@@ -171,16 +171,36 @@
...
+             <:link>
+             <AuLinkExternal
+               @skin="button"
+               href={{(worship-organisations-database-url)}}
+             >
+               Ga naar de eredienst organisaties (externe link)
```
Windvis, line 198: Same here
```suggestion
               Ga naar de eredienst organisaties databank (externe link)
```

Windvis (commented):
Some small copy remarks but should be good to go otherwise 👍.

app/templates/index.hbs
cecemel: I don't think it's called databank in the end...

Windvis: Yea, I didn't know about the issue. I assumed it was another DB app 😄.

Windvis (approved):

PR #258, opened 2023-01-12
Add a mail link to the helpdesk on the person search pages

Windvis:
This allows users to notify the helpdesk about data issues since the user can't correct this themselves.

The email looks like this in Apple mail:
image

### DIFF
--- a/app/components/shared/persoon/persoon-search-result.hbs
+++ b/app/components/shared/persoon/persoon-search-result.hbs
@@ -16,5 +16,8 @@
   </div>
   <div class="au-o-grid__item au-u-2-5 au-u-1-5@medium au-u-text-right">
     <AuButton {{on 'click' this.select}}>Selecteer</AuButton>
+    <AuHelpText>
+      Zie je een fout in de data? <AuLinkExternal href={{this.mailContent}}>Meld het ons!</AuLinkExternal>
+    </AuHelpText>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


Poltergeistz:
In thunderbird I don't see the body. I'm looking if it's something enabled by default from the app.

Windvis:
> cool didn't know this was possible. That's a standard?

I think so yea, the docs here mention an RFC 6068 at least.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_an_email_address

> In thunderbird I don't see the body. I'm looking if it's something enabled by default from the app.

Yea, I was afraid of that happening. I assume the most important client to test would be outlook, but I don't have that installed.

In any case, there is not much we can do about it other than using a form in the app and sending an actual email from the server. I also consider the body not showing a graceful degraduation. Users can still send an email, they just have to type more themselves :smile:.

Poltergeistz:
> > In thunderbird I don't see the body. I'm looking if it's something enabled by default from the app.
> 
> Yea, I was afraid of that happening. I assume the most important client to test would be outlook, but I don't have that installed.
> 
> In any case, there is not much we can do about it other than using a form in the app and sending an actual email from the server. I also consider the body not showing a graceful degraduation. Users can still send an email, they just have to type more themselves smile.

I'll install that and test to be sure. Gmail is working :+1:

Poltergeistz (approved):

PR #259, opened 2023-01-16
[Worship-modules] Fixing secondaryContactPoint is null error

Poltergeistz:
DL-4733

**Context :**
If we try to validate an empty address with no secondary phone number. The secondary phone number record is deleted _(this prevents the record to be saved empty in the database)_, but if we try again with an address it throws an error because it’s looking for that secondary contact point _(Secondary phone number)_ to be destroyed again when the secondary phone number is missing.

**Solved :**
By moving the "destroyRecord" logic in the final validation we make sure that the record is deleted once.

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -135,9 +135,6 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       if ((yield isValidPrimaryContact(contactPoint)) && this.model.isValid) {
         if (secondaryContactPoint.telefoon) {
           yield secondaryContactPoint.save();
-        } else {
-          // The secondary contact point is empty so we can remove it if it was ever persisted before
-          yield secondaryContactPoint.destroyRecord();
         }
 
         if (adres?.isNew || adres?.hasDirtyAttributes) {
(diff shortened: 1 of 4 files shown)


app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -196,6 +193,14 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
       this.model.isValid
     ) {
+      let secondaryContactPoint = yield this.selectedContact
+        .secondaryContactPoint;
+      if (secondaryContactPoint !== null) {
+        if (!secondaryContactPoint.telefoon) {
+          // The secondary contact point is empty so we can remove it if it was ever persisted before
+          yield secondaryContactPoint.destroyRecord();
+        }
+      }
```
Windvis, line 203: I think this change hides the real underlying issue. I think the issue is we are doing saves / destroys when the form isn't fully valid.

In this case, the address isn't valid, but we are still persisting other records. We should not do any of that as long as the form is not 100% valid.

https://github.com/lblod/frontend-loket/blob/81c764fd1c384e9f6ea958c9d0b3272ec11ac2c7/app/controllers/worship-ministers-management/minister/edit.js#L146-L150

That check would fail and early return which means the code isn't executed anymore.

I think the better fix is to add the `adres.isValid` check here as well: https://github.com/lblod/frontend-loket/blob/81c764fd1c384e9f6ea958c9d0b3272ec11ac2c7/app/controllers/worship-ministers-management/minister/edit.js#L138

That should have the same end result. The whole flow of that method can probably use some refactoring in the future though.

app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -196,6 +193,14 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
       this.model.contacts.length > 0 &&
       this.model.isValid
     ) {
+      let secondaryContactPoint = yield this.selectedContact
+        .secondaryContactPoint;
+      if (secondaryContactPoint !== null) {
+        if (!secondaryContactPoint.telefoon) {
+          // The secondary contact point is empty so we can remove it if it was ever persisted before
+          yield secondaryContactPoint.destroyRecord();
```
Windvis, line 201: I would also move it back to the previous location since that should no longer trigger the error now.

Windvis (approved):

PR #261, opened 2023-01-18
Allow users to preview a public service concept before creation

Windvis:
This adds a link to the public service concept details page so users can preview the contents before creating an instance.


Poltergeistz (approved):
Works as expected :+1:

PR #262, opened 2023-01-20
Make users aware of self-registration option

Riadabd:
Make users aware of the self-registration option on the loket page (Addresses DL-4735).


claire-lovisa (commented):
Hey ! Works very fine. A suggestion maybe, we use this library of components to make our applications uniform https://github.com/appuniversum/ember-appuniversum, and i think that the `AuLinkExtrernal` component would be a good fit for your links here ? It also comes with the `target` and `rel` attributes to open the link in a new tab automatically, which is a bit more user friendly :smile:

claire-lovisa (approved):

PR #266, opened 2023-01-23
Set up ember-template-imports

Windvis:
This allows us to create components in the new style that will be rolled out in future Ember versions.

It seems to work fine for the most part. ember-template-lint seems to throw some `no-implicit-this` false positives, but worst case we can disable that rule voor `.gjs` files.

I'll experiment with this some more before merging, but it would be nice if all new components would be implemented in the new style since it means less refactoring in the future. There are also some classic components that need to be converted to Glimmer components, and ideally it would be in this new syntax as well.

(The DataList refactor was simply a test to see if everything keeps working as expected. That shouldn't be part of the setup PR so I'll split it off before merging)


.eslintrc.js
```diff
@@ -12,11 +12,7 @@ module.exports = {
     },
   },
   plugins: ['ember'],
-  extends: [
-    'eslint:recommended',
-    'plugin:ember/recommended',
-    'plugin:prettier/recommended',
-  ],
+  extends: ['eslint:recommended', 'plugin:ember/recommended', 'prettier'],
```
NullVoxPopuli, line 15: You may want to remove prettier here as well

Windvis: This refers to [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) (not the plugin) which only disables rules that might conflict with prettier.

It's possible that there are no conflicts with the other configs, but I don't think adding it here is a bad thing?

Windvis:
Closing for now since it's still not officially recommended by the Ember team.

NullVoxPopuli:
not recommended doesn't mean don't use it ;) 

just has some rough edges here and there (like with linting)

PR #268, opened 2023-01-25
LPDC: Implement the basic "change notification" system

Windvis:
On the overview page we now show a label if the concept was changed. It's also possible to display only the services that have underlying changes to make it easier for the user.

On the details page we now show an alert if the concept is changed (with slightly different text based on the type of change) with a link to the concept details page, and a button to mark the service as reviewed.

**Testing**
This branch is needed: https://github.com/lblod/app-digitaal-loket/pull/320
The easiest way to test this is to add some temporary code to the public-services.details route model hook:

```js
  async model({ serviceId }) {
    const publicService = await loadPublicServiceDetails(this.store, serviceId);
   
    // "Updated" status
    let status = await this.store.findRecord('concept', '5a3168e2-f39b-4b5d-8638-29f935023c83');
    // or "Deleted" status
    // let status = yield this.store.findRecord('concept', 'cf22e8d1-23c3-45da-89bc-00826eaf23c3');
    publicService.reviewStatus = status;
    await publicService.save();
   
   //...
```
You can then browse to the details page of the service you want to give a certain state. Once the state is updated you can comment it again and refresh the page.

### DIFF
--- a/app/components/public-services/details-page.hbs
+++ b/app/components/public-services/details-page.hbs
@@ -40,7 +40,7 @@
   {{/if}}
 </AuBodyContainer>
 
-<AuToolbar @border="top" @size="large" as |Group|>
+<AuToolbar @border="top" @size="medium" as |Group|>
   <Group>
     <AuButtonGroup>
       {{#if @readOnly}}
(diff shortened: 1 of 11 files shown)


Poltergeistz (approved):
Works as expected

Windvis:
Whoops, just noticed that the "Concept bekijken" button isn't hidden yet when the concept is deleted 😄. Will fix in a follow-up PR.

PR #276, opened 2023-02-07
Replace the datepicker with the `<AuDateInput>` component

Windvis:
This makes it easier for users to type in the date which will speed up data entry, especially if dates are far away in the past or future.

### DIFF
--- a/app/components/leidinggevendenbeheer/functionaris-form.hbs
+++ b/app/components/leidinggevendenbeheer/functionaris-form.hbs
@@ -8,10 +8,11 @@
     <div class="au-o-grid au-o-grid--tiny">
       <div class="au-o-grid__item au-u-1-2 au-u-1-3@medium">
         <AuLabel for="date-picker-start">Start aanstellingsperiode</AuLabel>
-        <AuDatePicker
+        <AuDateInput
           @value={{@model.start}}
           @onChange={{fn this.handleDateChange "start"}}
-          @id="date-picker-start"
+          @width="block"
+          id="date-picker-start"
         />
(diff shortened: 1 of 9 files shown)


Poltergeistz (approved):
Everything works as expected :+1:

app/components/leidinggevendenbeheer/functionaris-form.hbs
```diff
@@ -8,21 +8,21 @@
     <div class="au-o-grid au-o-grid--tiny">
       <div class="au-o-grid__item au-u-1-2 au-u-1-3@medium">
         <AuLabel for="date-picker-start">Start aanstellingsperiode</AuLabel>
-        <AuDatePicker
+        <AuDateInput
           @value={{@model.start}}
           @onChange={{fn this.handleDateChange "start"}}
```
Windvis, line 13: I need to double check if this also doesn't need the `@width="block"` arg.

Poltergeistz: Sure, I'll check

Windvis: It seems the previous version was filling the width, so I'll add that here as well:

image

Poltergeistz: It works well without it. There is something I've noticed for small devices, if you don't provide the `block` arg the two input will overlap a bit. And for larger resolution they will be far from each other. 

So I think it's better for UX to add this argument.

Windvis:
Thanks for the review 👍

PR #281, opened 2023-02-21
Add productID to both index and add pages for public services

Riadabd:
(Addresses DL-4551) Allows productID to be visible for users in the index and add pages of the services module.

### DIFF
--- a/app/templates/public-services/add.hbs
+++ b/app/templates/public-services/add.hbs
@@ -44,7 +44,7 @@
         @currentSorting={{this.sort}}
         @label="Productnaam"
       />
-      <th>ProductID</th>
+      <th>IPDC Concept ID</th>
       <AuDataTableThSortable
         @field="type"
         @currentSorting={{this.sort}}
--- a/app/templates/public-services/add.hbs
+++ b/app/templates/public-services/add.hbs
@@ -44,6 +44,7 @@
         @currentSorting={{this.sort}}
         @label="Productnaam"
       />
+      <th>ProductID</th>
       <AuDataTableThSortable
         @field="type"
         @currentSorting={{this.sort}}
@@ -95,6 +96,7 @@
             {{{publicService.nameNl}}}
           </AuLink>
         </td>
+        <td>{{publicService.productId}}</td>
         <td>{{publicService.type.label}}</td>
         <td>
           {{#each publicService.targetAudiences as |targetAudience|}}
--- a/app/templates/public-services/details.hbs
+++ b/app/templates/public-services/details.hbs
@@ -25,7 +25,6 @@
                   <AuLink
                     @route="public-services.concept-details"
                     @model={{@model.publicService.concept.id}}
-                    @query={{hash preview=true}}
                   >
                     {{productId}}
                   </AuLink>
--- a/app/templates/public-services/details.hbs
+++ b/app/templates/public-services/details.hbs
@@ -20,13 +20,17 @@
         <Data>
           <:title>IPDC Concept ID</:title>
           <:content>
-            <AuLink
-              @route="public-services.concept-details"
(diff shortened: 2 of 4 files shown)


app/templates/public-services/add.hbs
```diff
@@ -44,6 +44,7 @@
         @currentSorting={{this.sort}}
         @label="Productnaam"
       />
+      <th>ProductID</th>
```
Windvis, line 47: I would make it sortable if possible. It should be as simple as mimicking the other sortable column components and modifying the arguments. 

Maybe it also makes sense to make this the default sort instead of the 'name', but I'll double check with Stefanie about that.

app/templates/public-services/index.hbs
```diff
@@ -56,6 +56,9 @@
       <th>
         Productnaam
       </th>
+      <th>
```
Windvis, line 59: According to the Jira issue we don't need to show it here, but on the details page.

app/templates/public-services/index.hbs
```diff
@@ -118,6 +121,7 @@
             </AuPill>
           {{/if}}
         </td>
+        <td>{{publicService.productId}}</td>
```
Windvis, line 124: Not super familiar with the data model, but shouldn't the data be accessed through the concept? (`{{publicService.concept.productId}}`)

Or does each instance also receive a copy of the product id value?

app/templates/public-services/add.hbs
Windvis, line 47: I also think the label should be `IPDC Concept ID`.

app/templates/public-services/index.hbs
Riadabd: Yes, you're right; I missed that we can use the conceptual service attributes for the instance ones as well. Thanks for catching it!

app/templates/public-services/details.hbs
```diff
@@ -17,6 +17,18 @@
...
     <DataList as |Data|>
+        <Data>
+          <:title>IPDC Concept ID</:title>
+          <:content>
+            <AuLink
+              @route="public-services.concept-details"
+              @model={{@model.publicService.concept.id}}
+              @query={{hash preview=true}}
+            >
+              <ValueWithPlaceholder @value={{@model.publicService.concept.productId}} />
```
Windvis, line 28: I think it should be the other way around. Pseudo:

```hbs
<ValueWithPlaceholder @value={{@model.publicService.concept.productId}}>
  <:value as |productId|>
     <AuLink>{{productId}}</AuLink>
  </:value>
</ValueWithPlaceholder>
```

The current version will always try to link even if there is no productId.

app/templates/public-services/add.hbs
Windvis: I would still make it sortable even if we don't make it the default sorted column. I could imagine that some users would want to list the ids in an ascending order.

app/templates/public-services/details.hbs
```diff
@@ -17,6 +17,22 @@
...
     <DataList as |Data|>
+        <Data>
+          <:title>IPDC Concept ID</:title>
+          <:content>
+            <ValueWithPlaceholder @value={{@model.publicService.concept.productId}}>
+                <:value as |productId|>
+                  <AuLink
+                    @route="public-services.concept-details"
+                    @model={{@model.publicService.concept.id}}
+                    @query={{hash preview=true}}
```
Windvis, line 28: I don't think we want to set this to true here. The preview option is more intended for the creation flow.

Windvis (approved):
Other than the preview remark this looks good to me. I'll try to test it with the backend branches ASAP (but probably tomorrow morning).

aatauil (approved):

PR #282, opened 2023-02-24
Update to @lblod/ember-acmidm-login v2

Windvis:
This replaces `@lblod/ember-acmidm-login` to v2. 

V2 has a couple of nice advantages:
- no torii dependency (unmaintained addon)
- no Appuniversum dependency. The addon is now completely style agnostic.
- no pop-ups, but real redirects so it will work in all browsers. (pop-ups where usually blocked on the switch flow).
- It allows us to start the login flow from an external website which is a requirement for the new website.

**To test this locally:**

- Generate ssl certificates following this guide: https://mainmatter.com/blog/2022/09/22/selfsigned-certificates-for-development/
- edit `/etc/hosts` so `loket.lblod.info` maps to your local machine
- run ember serve on part 443 and use the --ssl argument: `ember serve --port 443 --ssl --proxy https://your-backend` (I had to run this with sudo since the port can't be used as non-root on macOS).
- Log in through ACM/IDM (you need to have access to the ACM/IDM QA environment)


app/components/page-loader.hbs
```diff
@@ -0,0 +1,13 @@
+<div
+  class="au-o-box au-u-flex au-u-flex--column au-u-flex--vertical-center"
```
Windvis, line 2: I've based the implementation on the [Webuniversum loader example](https://overheid.vlaanderen.be/webuniversum/v3/documentation/atoms/vl-ui-loader).

app/routes/auth/callback.js
```diff
@@ -0,0 +1,26 @@
...
+  beforeModel() {
+    this.session.prohibitAuthentication('index');
+  }
+
+  async model({ code }) {
+    if (code) {
+      try {
+        await this.session.authenticate('authenticator:acm-idm', code);
+      } catch (error) {
+        throw new Error(
```
Windvis, line 17: I throw an error here to enter the error substate since we don't have a generic error page in loket.

app/routes/auth/logout.js
```diff
@@ -0,0 +1,29 @@
...
+  @service router;
+  @service session;
+
+  async beforeModel(transition) {
+    if (this.session.requireAuthentication(transition, 'login')) {
+      try {
+        let wasMockLoginSession = this.session.isMockLoginSession;
+        await this.session.invalidate();
+        let logoutUrl = wasMockLoginSession
+          ? this.router.urlFor('mock-login')
```
aatauil, line 15: Thanks for adding this! small change but big dev impact :smile:

aatauil (commented):
Code LGTM! Best to also wait for Sergey's usertest before merging

Windvis:
Just noticed that I still need to remove this sentence from the login template: https://github.com/lblod/frontend-loket/blob/4b8c21860276676c0feffb9b5239703b2cbd9cd7/app/templates/login.hbs#L16

app/services/session.js
```diff
@@ -1,24 +1,24 @@
...
-    if (logoutUrl.startsWith('http')) {
-      super.handleInvalidation(logoutUrl);
-    } else {
-      warn('Incorrect logout URL configured', {
-        id: 'session-invalidation-failure',
-      });
-    }
+    // We don't want the default redirecting logic of the base class since we handle this ourselves in other places already.
+    // We can't do the logic here since we don't know which authenticator did the invalidation and we don't receive the arguments that are passed to `.invalidate` either.
+    // This is needed to be able to support both normal logouts, switch logouts (and as a bonus,also mock logouts).
```
Windvis, line 22: This is not an ideal solution, but we want to do different actions depending on the used authenticator and also different actions depending on the way we want to log out. I couldn't find a way to get this information in this method so I've handled it at the places where the logout is started.

I think this is a limitation in the ESA API, but I would have to think about it some more.

Asergey91 (approved):
did some minor testing and skimmed the code, looks good

PR #283, opened 2023-02-27
[Eredienst mandatenbeheer] Prefill the endDate based on the expectedEndDate

Poltergeistz:
DL-4380

- Prefills the endDate from the expectedEndDate when a mandate is selected.
- Adds warning messages when the endDate is set by the user.

### DIFF
--- a/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
+++ b/app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
@@ -9,10 +9,12 @@ import {
   findPrimaryContactPoint,
   isValidPrimaryContact,
 } from 'frontend-loket/models/contact-punt';
-import { setExpectedEndDate } from 'frontend-loket/utils/eredienst-mandatenbeheer';
+import {
+  handlePrefillEndDate,
+  setExpectedEndDate,
+} from 'frontend-loket/utils/eredienst-mandatenbeheer';
 import { validateMandaat } from 'frontend-loket/models/worship-mandatee';
 import { combineFullAddress, isValidAdres } from 'frontend-loket/models/adres';
-import moment from 'moment';
 export default class EredienstMandatenbeheerMandatarisEditController extends Controller {
   @service currentSession;
   @service store;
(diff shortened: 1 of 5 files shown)


app/controllers/eredienst-mandatenbeheer/mandataris/edit.js
```diff
@@ -28,15 +33,32 @@ export default class EredienstMandatenbeheerMandatarisEditController extends Con
...
-    setExpectedEndDate(this.store, this.model, mandaat);
+    await setExpectedEndDate(this.store, this.model, mandaat);
+    const endDate = this.userInputEndDate || this.model.expectedEndDate;
+    const { userInputEndDate, warningMessages } = handlePrefillEndDate(
+      this.userInputEndDate,
+      this.model,
+      endDate
+    );
+    this.userInputEndDate = userInputEndDate;
+    this.warningMessages = warningMessages;
```
Windvis, line 46: I think most of this can be simplified by simply checking the previous date with the previous expected end date. If they were different we know it was user modified.

I discussed this again with Saška, and we can simplify it even more by always overriding the end date with the new expectedEndDate, and if the new expectedEndDate differs from the previous end date, we should show a warning.

Windvis (changes requested):
I was testing things out locally for the discussion with Saska, so I'll just push the update that also fixes the comments I made here.

app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
```diff
@@ -86,30 +86,32 @@
...
-        {{/if}}
-        {{#if showWarning}}
-          <AuHelpText @warning={{true}}>
-            {{this.warningMessages.userInputEndDateMessage}}
-          </AuHelpText>
-        {{/if}}
-        {{/let}}
+        {{#let
+          (if this.model.errors.einde true false)
+          (if this.warningMessages.einde true false)
```
Windvis, line 91: You can use a let to output multiple values, no need to wrap 2 lets (unless you need a value from the outer one).

Poltergeistz:
> I was testing things out locally for the discussion with Saska, so I'll just push the update that also fixes the comments I made here.

:+1: Thanks

app/utils/eredienst-mandatenbeheer.js
```diff
@@ -21,3 +22,39 @@ export async function setExpectedEndDate(store, mandataris, mandaat) {
...
+ * @returns Object containing `worshipMandatee` `userInputEndDate` `warningMessages`
+ */
+export function handlePrefillEndDate(
+  userInputEndDate,
+  worshipMandatee,
+  endDate
+) {
+  if (
+    worshipMandatee.expectedEndDate &&
+    moment(worshipMandatee.expectedEndDate).isSame(moment(endDate))
```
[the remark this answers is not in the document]
Poltergeistz: Ah okay, sorry didn't want to add extra `moment` issues.

PR #288, opened 2023-03-15
Add a build-time feature flag to all "controle" logic

Windvis:
This uses the [embroider macros](https://github.com/embroider-build/embroider/tree/main/packages/macros) to add a flag that allows us to toggle the controle functionality at build time.

This allows us to merge the 'batch-edit' branch back into the main branch which makes maintenance and releases easier.

`npm run start:controle` can be used to run the controle version of the app. A new build-pipeline will be used to build the controle image. The -prod version is dropped since it seems we didn't use that in production.

I used [this diff view](https://github.com/lblod/frontend-loket/compare/development...9146a21440f59533a77459b5cfb0ee7921126c75) as the reference.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -102,6 +102,26 @@ trigger:
   event:
     - tag
 ---
+kind: pipeline
+type: docker
+name: release-controle
+steps:
+- name: push-tagged-build
+  image: plugins/docker
+  settings:
+    repo: ${DRONE_REPO}
+    tags: ${DRONE_TAG##v}-controle
+    username:
+      from_secret: docker_username
+    password:
+      from_secret: docker_password
+    purge: true
+    build_args:
+      - CONTROLE=true
+trigger:
+  event:
+    - tag
+---
 kind: secret
 name: docker_username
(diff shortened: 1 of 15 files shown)


.drone.yml
```diff
@@ -102,6 +102,26 @@ trigger:
...
+  settings:
+    repo: ${DRONE_REPO}
+    tags: ${DRONE_TAG##v}-controle
+    username:
+      from_secret: docker_username
+    password:
+      from_secret: docker_password
+    purge: true
+    build_args:
+      - CONTROLE=true
```
Windvis, line 120: I didn't test this yet, but in theory this should override the default ARG value set in the Dockerfile.

app/controllers/supervision/submissions/index.js
```diff
@@ -10,11 +10,6 @@ export default class SupervisionSubmissionsIndexController extends Controller {
   size = 20;
   sort = 'status.label,-sent-date,-modified';
 
-  @action
-  lookAt(submission) {
```
Windvis, line 14: This wasn't used anywhere so I removed it.

PR #290, opened 2023-03-23
Implement the concept label and filtering functionality

Windvis:
This:
- adds a new sidebar with filters to the "create instance" and "link concept" flows
- shows "nieuw" and "toegevoegd" labels next to the concept names depending on the state
- adds some logic to manage the state when linking and unlinking concepts

This depends on:
- https://github.com/lblod/app-digitaal-loket/pull/412
- https://github.com/lblod/lpdc-management-service/pull/33

DL-4956


Windvis:
It currently looks like this:

image

Windvis:
We need to do the same changes to the route that was introduced in #297

Riadabd (approved):
All works as expected.

PR #291, opened 2023-03-23
[Eredienst mandatenbeheer] Begin and End date should be limited to the period bounds

Poltergeistz:
DL-4389

**Adding :** 
- Warning messages  below inputs `Startdatum` and `Einddatum` when user cross period bounds limits for New and Edit pages.
- `formatDate` utility function that format the date with the following date format DD-MM-YYYY so we wont have to use moment.format. Note for later : Maybe this function can be used as an helper too so we can remove moment from templates as well.

Screenshot 2023-03-23 at 13-54-14 Bewerk Mandataris Lorem Ipsum Mandatenbeheer Loket voor lokale besturen

**Fixed :** 
- Case where selecting `Bestuurslid (van rechtswege) van het bestuur van de eredienst` mandate shows 'De einddatum werd automatisch aangepast. Gelieve de einddatum te controleren.' then selecting right after an other mandate it showed the same message. Exceptions for `Bestuurslid van het bestuur van de eredienst (Grote Helft)` mandate who shows 'De einddatum werd automatisch aangepast naar de nieuwe geplande einddatum. Gelieve de einddatum te controleren.'

Now it shows the right message for each cases.

### DIFF
--- a/app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
+++ b/app/templates/eredienst-mandatenbeheer/mandataris/edit.hbs
@@ -80,7 +80,7 @@
             @error={{showError}}
             @warning={{showWarning}}
             @width="block"
-            @id="mandate-correction-start-date"
+            id="mandate-correction-start-date"
           />
           {{#if showError}}
             <AuHelpText @error="true">
@@ -121,7 +121,7 @@
             @error={{showError}}
             @warning={{showWarning}}
             @width="block"
-            @id="mandate-correction-end-date"
+            id="mandate-correction-end-date"
           />
           {{#if showError}}
             <AuHelpText @error={{true}}>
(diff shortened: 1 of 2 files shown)


app/utils/eredienst-mandatenbeheer.js
```diff
@@ -64,3 +75,86 @@ export function isLifetimeBoardPosition(boardPosition) {
...
+    case 'Secretaris van het bestuur van de eredienst':
+      activeTimePeriodeLimitStart = activeTimePeriodes[0].bindingStart;
+      activeTimePeriodeLimitEnd = activeTimePeriodes[0].bindingEinde;
+      break;
+    case 'Bestuurslid van het bestuur van de eredienst (Grote Helft)':
+    case 'Bestuurslid van het bestuur van de eredienst (Kleine Helft)':
+      activeTimePeriodeLimitStart = activeTimePeriodes[0].bindingStart;
+      activeTimePeriodeLimitEnd = activeTimePeriodes[1].bindingEinde;
+      break;
+    case 'Bestuurslid (van rechtswege) van het bestuur van de eredienst':
```
Windvis, line 138: I would check this early in the util (isLifetimeBoardPosition) and do an early return. The checks below work in that case because comparing a date with undefined always returns false but it feels a bit icky to depend on that.

PR #292, opened 2023-03-24
Merge "batch-edit" back into the main branch setup

Windvis:
with #288 merged we can now safely merge the batch edit branch back into the main branching setup.

Running the "controle" version of the app can be done by executing `npm run start:controle ...` or by setting the `CONTROLE` environment variable to `"true"` before starting the app.

New releases of the controle image will now automatically be created when tagging a new release.


Windvis:
I wonder If I should rebase this branch and squash all merge commits into one. This will add a lot of noise to the history 😄.

PR #294, opened 2023-04-06
Add a `UrbanRenewal::FinancingTotals` custom field component

Windvis:
This field shows a "totals" summary based on the entries in the financing table.  

The styling looks a bit wonky, but that's because of the conflicting table styles in _shame.scss. We still need to convert the old custom tables to the new AuTable component before we can delete those.

image


aatauil (approved):

app/routes/subsidy/applications/edit.js
```diff
@@ -95,6 +96,10 @@ export default class SubsidyApplicationsEditRoute extends Route {
         edit: AccountabilityTableEditComponent,
         show: AccountabilityTableShowComponent,
       },
+      {
+        displayType: 'http://lblod.data.gift/display-types/financingTotals',
```
Windvis, line 100: I guess it would make sense to prefix the display type with 'urbanRenewal' since it's specifically made for that table.

aatauil: might possibly collide in future yeah. ns with urbanRenewal would be nice :)

PR #296, opened 2023-04-17
Allow users to unlink an instance from a concept

Windvis:
DL-5054

Best to be reviewed by hiding whitespace changes:
image

To test:
- create a new public service instance by selecting a concept in the list
- the unlink button should be visible
- clicking it should ask for confirmation
- confirming should remove the link to the concept (and in a future PR it will then be possible to link to a new concept again)
- Unlinking should not be possible if the status is "published", so users will have to open an instance for editing before the button appears again.

### DIFF
--- a/app/models/public-service.js
+++ b/app/models/public-service.js
@@ -34,3 +34,9 @@ export function isConceptUpdated(reviewStatus) {
 export function isConceptDeleted(reviewStatus) {
   return reviewStatus?.uri === REVIEW_STATUS.DELETED;
 }
+
+export function hasConcept(publicService) {
+  // This assumes the relationship was already loaded. Either by using includes, or by resolving the promise.
+  // Otherwise this will always return false
+  return Boolean(publicService.belongsTo('concept').id());
+}
--- a/app/templates/public-services/details.hbs
+++ b/app/templates/public-services/details.hbs
@@ -16,81 +16,125 @@
   as |Group|
 >
   <Group class="au-u-margin-bottom-small">
-    <DataList as |Data|>
+    <div>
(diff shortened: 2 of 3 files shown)


app/templates/public-services/details.hbs
```diff
@@ -16,81 +16,126 @@
...
+          @size="small"
+          @closable={{this.unlinkConcept.isIdle}}
+          @onClose={{this.hideUnlinkWarning}}
+          class="au-u-margin-top au-u-max-width-xsmall"
+        >
+          {{! TODO: Add the correct copy, or remove it and only show the title }}
+          {{! <p>
+            Uitleg koppelen + gevolgen ontkoppeling uitleggen? uitleg wanneer
+            wel juist is om los te koppelen?
+          </p> }}
```
Windvis, line 121: Waiting for input from Geertrui. Worst case we don't display a message.

app/models/public-service.js
```diff
@@ -14,6 +14,10 @@ export default class PublicServiceModel extends ConceptualPublicServiceModel {
       'http://lblod.data.gift/concepts/9bd8d86d-bb10-4456-a84e-91e9507c374c'
     );
   }
+
+  get hasConcept() {
+    return Boolean(this.belongsTo('concept').id());
```
Windvis, line 19: Hmm, it seems this isn't safe to depend on. If the backend doesn't include the id in the relationship config (but only a link) this will return false while it _does_ have a concept if the link is fetched.

benjay10 (approved):
Not an expert here, but this seems to work as intended.
Tested on my local setup.

PR #298, opened 2023-04-21
Fix the rich text editor field validations

Windvis:
It seems one of the say-editor updates caused a change in the `htmlContent` value when the editor is (visually) empty. Before the update it used to be an empty string, but now it returns an empty `<p>` tag instead. This also has the side effect that the "required" validation no longer works as expected since the field has content (even if it is not visible). We now check the `textContent` to determine if the field is empty or not.

We now also set the `hasBeenFocused` to true when the field is focused, so the validation errors will be shown earlier, in line with the other form fields.


Windvis:
Indeed, I just wanted a quick fix. Do you want me to add a todo in the code?

cecemel:
> Indeed, I just wanted a quick fix. Do you want me to add a todo in the code?

if it helps us not to forget!

PR #303, opened 2023-05-02
Disable autofill on start and end date fields when creating mandataris

Riadabd:
(Addresses DL-5081) Disables auto-fill on start and end date fields when creating a mandate which was causing issues with the validation code of the `Startdatum` field.

I checked the mandataris edit page to see if auto-fill shows up, but I did not see any browser autopilot window when setting/changing the start and end dates.


Windvis:
I think we should just do this in Appuniversum actually. We might run into it again in the future otherwise.

Riadabd:
That sounds better indeed. I'll close this PR.

PR #308, opened 2023-05-08
Remove the -prod image builds

Windvis:
The difference between the QA and PRD builds is very small but it requires more manual actions when deploying. We now convert this to a single image by retrieving the ACMIDM config from the server.

The test selectors are now no longer stripped in production builds, but that doesn't matter that much for the final build size.


config/environment.js
```diff
@@ -27,12 +27,12 @@ module.exports = function (environment) {
       includeTimezone: 'all',
     },
     acmidm: {
-      clientId: 'a2c0d6ea-01b4-4f68-920b-10834a943c27',
+      clientId: '{{ACMIDM_CLIENT_ID}}',
```
Windvis, line 30: This means we need to set the values on the server now when deploying the new frontend image.

cecemel (approved):
thanks! checked it locally seems ok. You do the release boilerplate?

PR #309, opened 2023-05-16
set default status for worship mandatee to EFFECTIEF

aatauil:
Whenever we create a new worship mandatee. We want to set the status to EFFECTIEF by default. The status is shown in OP but currently its always empty for new mandatees

test instructions: https://github.com/lblod/app-digitaal-loket/pull/416


aatauil:
Ah, yeah forgot about that step. Havent touched frontend for a while

PR #310, opened 2023-05-24
Add support for handling external subsidies

Riadabd:
(Addresses DL-4930 and DL-5120) Adjusts frontend to handle support for external subsidies.

This is to be tested with the loket PR (https://github.com/lblod/app-digitaal-loket/pull/421).

### DIFF
--- a/app/models/subsidy-measure-offer-series.js
+++ b/app/models/subsidy-measure-offer-series.js
@@ -24,27 +24,35 @@ export default class SubsidyMeasureOfferSeriesModel extends Model {
   /**
    * Returns if the series can be processed externally
    * - when it has a reference to an outside form
-   * - when the subsidyProceduralStep of the firstApplicationStep is of type ExternallyProcessed
+   * - when the subsidyProceduralStep of the firstApplicationStep is of type "Externally Processed"
    * @returns boolean
    */
-  get isExternallyProcessed() {
-    const isReplacedBy = this.activeApplicationFlow.get(
-      'firstApplicationStep.isReplacedBy'
-    );
-    const isExternallyProcessed = this.activeApplicationFlow.get(
-      'firstApplicationStep.subsidyProceduralStep.isExternallyProcessed'
-    );
+  isExternallyProcessed() {
+    const activeApplicationFlow = this.belongsTo(
+      'activeApplicationFlow'
+    ).value();
+    const firstApplicationStep = activeApplicationFlow
+      .belongsTo('firstApplicationStep')
+      .value();
+    const isReplacedBy = firstApplicationStep.isReplacedBy;
+    const isExternallyProcessed = firstApplicationStep
+      .belongsTo('subsidyProceduralStep')
+      .value().isExternallyProcessed;
+
     if (isReplacedBy && !isExternallyProcessed) {
       console.warn(
         'found a link to an external processes, but step is not marked for external processing.'
       );
+
       return false;
     } else if (!isReplacedBy && isExternallyProcessed) {
       console.warn(
         'found no link to an external processes, but step is marked for external processing.'
       );
+
       return true;
     }
+
     return isReplacedBy && isExternallyProcessed;
   }
 }
--- a/app/models/subsidy-measure-offer-series.js
+++ b/app/models/subsidy-measure-offer-series.js
@@ -20,4 +20,31 @@ export default class SubsidyMeasureOfferSeriesModel extends Model {
     const end = moment(this.period.get('end')).format('YYYY');
     return `${begin} – ${end}`;
   }
+
+  /**
+   * Returns if the series can be processed externally
+   * - when it has a reference to an outside form
+   * - when the subsidyProceduralStep of the firstApplicationStep is of type ExternallyProcessed
+   * @returns boolean
+   */
+  get isExternallyProcessed() {
+    const isReplacedBy = this.activeApplicationFlow.get(
+      'firstApplicationStep.isReplacedBy'
+    );
+    const isExternallyProcessed = this.activeApplicationFlow.get(
+      'firstApplicationStep.subsidyProceduralStep.isExternallyProcessed'
+    );
+    if (isReplacedBy && !isExternallyProcessed) {
+      console.warn(
+        'found a link to an external processes, but step is not marked for external processing.'
+      );
+      return false;
+    } else if (!isReplacedBy && isExternallyProcessed) {
+      console.warn(
+        'found no link to an external processes, but step is marked for external processing.'
+      );
+      return true;
+    }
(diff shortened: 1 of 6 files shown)


app/routes/subsidy/applications/new.js
```diff
@@ -29,6 +29,12 @@ export default class SubsidyApplicationsNewRoute extends Route {
       this.router.transitionTo('subsidy.applications.available-subsidies');
     }
 
+    if (transition.data.series.isExternallyProcessed) {
+      // TODO: Show a warning / error page
+      console.warn('This subsidy application should be processed externally');
```
Windvis, line 34: I'm not sure this is needed? The only way to get in this state is if someone manually edits the url, so I think we can just catch this silently and do the redirect?

app/templates/subsidy/applications/available-subsidies.hbs
```diff
@@ -45,32 +67,66 @@
...
         {{series.label}}
         <AuHelpText @skin="secondary" class="au-u-margin-top-none">
-            {{#if (or series.period.begin series.period.end)}}
-              {{moment-format series.period.begin "DD MMMM YYYY"}}
-              &#8211;
-              {{moment-format series.period.end "DD MMMM YYYY"}}
-            {{else}}
-              N.V.T
-           {{/if}}
+          {{#if (and series.period.begin series.period.end)}}
```
Windvis, line 70: I wouldn't change that here, maybe it was intentional to show it like that? I wouldn't include it as part of this PR at least, even if it is a bugfix.

app/templates/subsidy/applications/available-subsidies.hbs
```diff
@@ -45,32 +67,66 @@
...
           Externe link, klik rechter muisknop om te openen
         </AuHelpText>
       </td>
       <td>{{moment-format series.deadline.end "DD MMMM YYYY"}}</td>
       <td>
-        <LinkTo @route="subsidy.applications.new" @query={{hash seriesId=series.id}} class="au-c-link">
-          Start aanvraag
-        </LinkTo>
+        {{#if series.isExternallyProcessed}}
+          <a
```
Windvis, line 104: Appuniversum now includes a component for this use-case: https://appuniversum.github.io/ember-appuniversum/?path=/story/components-links-aulinkexternal--primary

We just need to do a sweep across the app in a separate PR and replace all the external links where possible, but it's not super high prio. But new code should ideally use the component already.

Windvis (commented):
Some remarks about the implementation, but everything works as expected 👍.

app/templates/subsidy/applications/available-subsidies.hbs
```diff
@@ -45,32 +67,66 @@
...
+            Start aanvraag
+          </a>
+          <AuHelpText @skin="secondary" class="au-u-margin-top-none">
+            <AuIcon
+              @icon="export"
+              @alignment="left"
+              @size="default"
+              @ariaHidden={{true}}
+            />
+            Externe afhandeling
```
Windvis, line 119: Nitpick, I noticed the extra white space is visible so I would either remove it:
```suggestion
            />Externe afhandeling
```

Or remove the `@alignment` argument of the AuIcon component and depend on the whitespace for the spacing.

app/templates/subsidy/applications/available-subsidies.hbs
Riadabd: This wasn't a bug fix, but more so a way to allow the frontend to display something like `Start Date - N.V.T` rather than `Start Date - `. With the "or" in place, the if-branch would always execute since there's a start date for these external subsidies but no end date.

Windvis: Oh, I assumed there might also have been cases where the start date wasn't set and it would display as `- End date`, and in this setup that case is no longer supported. I'm not sure if that can happen in practice though (I'm not familiar with the data model), so maybe it's fine.

You could also do something like this so you don't need to repeat the formatting code:
```hbs
{{#if (or series.period.begin series.period.end)}}
  {{if series.period.begin (moment-format series.period.begin "DD MMMM YYYY") "N.V.T"}}
  &#8211;
  {{if series.period.end (moment-format series.period.end "DD MMMM YYYY") "N.V.T"}}
{{else}}
  N.V.T
{{/if}}
```

app/models/subsidy-measure-offer-series.js
```diff
@@ -24,27 +24,35 @@ export default class SubsidyMeasureOfferSeriesModel extends Model {
...
    * @returns boolean
    */
-  get isExternallyProcessed() {
-    const isReplacedBy = this.activeApplicationFlow.get(
-      'firstApplicationStep.isReplacedBy'
-    );
-    const isExternallyProcessed = this.activeApplicationFlow.get(
-      'firstApplicationStep.subsidyProceduralStep.isExternallyProcessed'
-    );
+  isExternallyProcessed() {
```
Windvis, line 30: It needs to be a getter for this to work. Alternatively it can be turned into a helper that receives the series record as an argument but that would still need other code changes.

Riadabd: I think I'm missing the link as to why it needs to be a getter. As it stands, it's working "by mistake"?

Windvis: Because the code expects it to be a getter (or property).

https://github.com/lblod/frontend-loket/pull/310/files#diff-170fddb1bb66da16f97a268b055af67ef8801c2e7de3fd9d292d1736831122a6R108

This check will always be true. It doesn't call the method, it converts it to boolean and function references are truthy. So it works for the cases that have external links, but all the others will also show the external link version even though they don't have one.

Riadabd: I included `active-application-flow.first-application-step.subsidy-procedural-step` without the `.period` so that the relationship fetching works; there is an increased processing time, but it's within reason I think. I'll push the commit for now to see if you find it acceptable.

Windvis: Yea, that should be fine, especially with the faster mu-auth / mu-resource 👍.

Riadabd:
The active subsidy filter relies on `:or:` operations, which are only present in `v1.22+` of `mu-cl-resources`. Merging this PR before resources is bumped will cause breakage.

Riadabd:
The `:or:` operations have been reverted for now, and temporary end dates have been added to the two external subsidies. I tested again on `v1.21.1` of resources to be on the safe side, and everything is displayed correctly.

Riadabd:
All of Marie's comments have been taken care of, and the PR now "hides" the dates for external subsidies and is working as expected.

Windvis (approved):

PR #312, opened 2023-06-05
Replace links to the old website with the new equivalents

Windvis:
With the go-live of the the new website, all old `https://lokaalbestuur.vlaanderen.be` links redirect to a generic `https://www.vlaanderen.be/lokaal-bestuur` page. This updates these links with their equivalents on the new website.


app/templates/subsidy/applications/edit/step/edit.hbs
```diff
@@ -127,7 +125,7 @@
                       Meer informatie over de subsidieaanvragen vindt u
                       <a
```
Windvis, line 126: Woops, this should also be replaced with AuLinkExternal.

PR #317, opened 2023-06-13
Disable the submit and save buttons if the form isn't loaded

Windvis:
This prevents an edge case where an exception is thrown if the user clicks the submit button before the form is loaded.


app/components/public-services/details-page.js
```diff
@@ -156,6 +172,10 @@ export default class PublicServicesDetailsPageComponent extends Component {
 
   @dropTask
   *saveSemanticForm() {
+    if (!this.canSave) {
+      return;
+    }
```
Windvis, line 177: Whoops, it seems I didn't properly test the flow again after adding this. This always fails since the task is running.

PR #319, opened 2023-06-19
Add a `lpdc-external` feature flag

Windvis:
This flag allows us to disable the LPDC module from the server. Once the new lpdc app is deployed to production we can remove the flag and all the code that is no longer needed.

Servers can enable this feature by adding the following config to the docker-compose.overrides.yml file:

```yml
services:
  frontend:
    environment:
      EMBER_FEATURE_LPDC_EXTERNAL: "true"
      EMBER_LPDC_URL: "https://ipdc.lokaalbestuur.lblod.info" # not the real url
```

To test, add temporary values for these in the `config/environment.js` file.

### DIFF
--- a/app/routes/lpdc-external-redirect.js
+++ b/app/routes/lpdc-external-redirect.js
@@ -1,0 +1,23 @@
+import Route from '@ember/routing/route';
+import config from 'frontend-loket/config/environment';
+import { inject as service } from '@ember/service';
+
+export default class LpdcExternalRedirectRoute extends Route {
+  @service router;
+
+  model(params, transition) {
+    if (!config.lpdcUrl.startsWith('{{')) {
+      const lpdcUrl = new URL(config.lpdcUrl);
+
+      if (transition.to.localName === 'with-path') {
+        const { path } = transition.to.params;
+        lpdcUrl.pathname = path;
+      }
+
+      window.location.replace(lpdcUrl);
+    } else {
+      // The URL isn't configured so we redirect to the index page instead.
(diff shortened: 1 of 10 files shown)


app/routes/lpdc-external-redirect.js
```diff
@@ -0,0 +1,23 @@
...
+  @service router;
+
+  model(params, transition) {
+    if (!config.lpdcUrl.startsWith('{{')) {
+      const lpdcUrl = new URL(config.lpdcUrl);
+
+      if (transition.to.localName === 'with-path') {
+        const { path } = transition.to.params;
+        lpdcUrl.pathname = path;
+      }
```
Windvis, line 15: This assumes that the new environment won't use the `/producten-en-dienstencatalogus` subpath.

So `https://loket.lblod.info/producten-en-dienstencatalogus/1234` would become `https://ipdc.loket.lblod.info/1234`.

It's possible this needs to be tweaked if the implementation is different. I could also keep it simple and always redirect to the toplevel domain without subpaths?

app/routes/lpdc-external-redirect.js
```diff
@@ -0,0 +1,23 @@
...
+
+      window.location.replace(lpdcUrl);
+    } else {
+      // The URL isn't configured so we redirect to the index page instead.
+      this.router.replaceWith('index');
```
Windvis, line 20: This case _shouldn't_ happen since it doesn't make sense to enable the feature without also configuring the url, so maybe it's overkill to do this.

PR #324, opened 2023-08-16
Add sites overview

andreo141:


lagartoverde (approved):
Looks good, I see some components added that I'm not sure if we need them like the secured-area, I assume you didn't add them, and Deniss did, but as we are going to switch apps (I will talk in the standup about this) it doesn't really matter, you can remove the tests as they are empty and you can merge into the epic

andreo141:
The securedarea components i copied from the OP template. I then took a look at the template in loket with dennis and we decided to leave it out because we didn't really knew what the securedarea was doing.

PR #330, opened 2023-08-24
Resolve ember v4 deprecations

Windvis:
This fixes all the deprecations I came across and also fixed some bugs that I missed in the upgrade PRs.

I believe I did a full sweep of all the codepaths in all the modules.. but I still might have missed things.


app/components/mandatenbeheer/mandaat-bestuursorganen-selector.js
```diff
@@ -9,7 +9,8 @@ export default Component.extend({
   _mandaat: oneWay('mandaat'),
   bestuursorganen: null,
 
-  async didReceiveAttrs() {
+  async init() {
```
Windvis, line 12: The component is conditionally rendered when the "mode" changes, so the didReceiveAttrs hook isn't needed. (This also makes it easy to convert it to a Glimmer component.)

PR #331, opened 2023-08-31
Fix an issue when reopening a supervision submission

Windvis:
This import doesn't exist (anymore?) so reopening fails. This only affected the controle environment so it went unnoticed.

### DIFF
--- a/app/controllers/supervision/submissions/index.js
+++ b/app/controllers/supervision/submissions/index.js
@@ -1,28 +1,34 @@
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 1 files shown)


app/controllers/supervision/submissions/index.js
```diff
@@ -17,7 +17,7 @@ export default class SupervisionSubmissionsIndexController extends Controller {
     );
     if (hasAcknowledged) {
       this.store
-        .findRecord('submission-document-status', CONCEPT_STATUS_UUID)
+        .findRecord('submission-document-status', CONCEPT_STATUS)
```
Windvis, line 20: Whoops, wrong fix, I didn't notice these constants were URLs 😄.

PR #332, opened 2023-09-01
Embroider v3 setup

Windvis:
Experiment to see how much work it is to get the app build under Embroider.

Targets the ember-v4-update branch for now.

There are some blockers caused by the rdfa-editor addon: https://github.com/lblod/ember-rdfa-editor/issues/945 Once those are resolved there might still be runtime issues, so extensive testing is still needed.


Windvis:
Once all the rdfa-editor fixes are released we can update and remove the patches.

PR #336, opened 2023-09-15
Remove the LPDC module

Windvis:
The cut-over to the separate app has been completed so we no longer need the code.

On hold until we receive the confirmation that we can remove the code.


app/templates/login.hbs
```diff
@@ -291,59 +329,71 @@
               </AuCard>
             </div>
 
-            {{#if (not (is-feature-enabled "lpdc-external"))}}
```
Windvis, line 294: I previously hid this when the feature was enabled to make it consistent with the other "external apps", but I'm not sure if we should?

app/routes/public-services.js
```diff
@@ -1,42 +0,0 @@
...
-    if (!this.currentSession.canAccessPublicServices)
-      this.router.transitionTo('index');
-  }
-
-  registerCustomFormFields() {
-    registerFormFields([
-      {
-        displayType: 'http://lblod.data.gift/display-types/richText',
-        edit: RichTextEditor,
-      },
```
Windvis, line 31: This means we can delete the component as well, since no other module is using it.

app/services/current-session.js
```diff
@@ -132,13 +132,6 @@ export default class CurrentSessionService extends Service {
   }
 
   get canAccessPublicServices() {
```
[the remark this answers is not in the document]
Windvis: Nevermind, not all users can redirect to the external app.

app/routes/public-services.js
Windvis: I'll do that in a separate PR to make it easier to roll back if needed.

PR #339, opened 2023-09-21
Add a conversation creation route

Windvis:
This will allow users to create new test conversations. The route is only accessible from the "controle" environment.

To test:
- run `npm run start:controle -- --proxy https://dev.loket.lblod.info`
- go to the berichtencentrum module
- click the new "nieuwe conversatie" button
- create a new conversation
- the edit page should now also show the conversation uri in an accordion

All of this should only be possible in the controle environment. The regular loket build should hide everything.


app/models/bericht.js
```diff
@@ -1,5 +1,11 @@
 import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 
+export const COMMUNICATION_TYPES = {
```
Windvis, line 3: These are the only 3 options I found (without looking in the DB), let me know if more need to be added.

app/models/bericht.js
```diff
@@ -1,5 +1,11 @@
...
+  KENNISGEVING: 'Kennisgeving toezichtsbeslissing',
+  OPVRAGING: 'Opvraging',
+  REACTIE: 'Reactie',
```
Windvis, line 6: Not sure if this needs to be an option in the select.

app/templates/berichtencentrum/berichten/new.hbs
```diff
@@ -0,0 +1,105 @@
...
+        </AuFormRow>
+
+        <AuFormRow>
+          <AuFieldset class="au-u-1-1" as |f|>
+            <f.legend @skin="6">
+              Bericht als bijlage
+            </f.legend>
+            <f.content class="au-u-1-1">
+              <div class="au-u-1-1">
+                {{#if @model.formData.file}}
```
Windvis, line 68: At the moment we allow only a single file upload. Should more be allowed?

app/controllers/berichtencentrum/berichten/new.js
```diff
@@ -0,0 +1,70 @@
...
+export default class BerichtencentrumBerichtenNewController extends Controller {
+  @service currentSession;
+  @service router;
+  @service store;
+
+  communicationTypes = Object.values(COMMUNICATION_TYPES);
+
+  get hasNeededData() {
+    const { dossiernummer, type, description, file } = this.model.formData;
+    return !!dossiernummer && !!type && !!description && !!file;
```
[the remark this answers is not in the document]
cecemel: no that's good for now

app/components/berichtencentrum/conversatie-view-header.hbs
```diff
@@ -23,10 +29,36 @@
...
+          worden gebruikt:
+          <AuPill>{{@conversatie.uri}}</AuPill>
+        </p>
+        <p>
+          Meer informatie kan je terugvinden in de
+          <AuLinkExternal
+            href="https://lblod.github.io/pages-vendors/#/docs/berichtencentrum-api"
+          >Berichtencentrum API documentatie</AuLinkExternal>.
+        </p>
+      </AuContent>
```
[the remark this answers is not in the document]
cecemel: that's ok for now!

app/models/bericht.js
cecemel: I don't remember others being, and since we're using it for testing

app/models/bericht.js
cecemel: For completeness it's good

app/templates/berichtencentrum/berichten/new.hbs
cecemel: This would be useful i think!

app/controllers/berichtencentrum/berichten/new.js
```diff
@@ -0,0 +1,70 @@
...
+          'http://data.lblod.info/id/bestuurseenheden/141d9d6b-54af-4d17-b313-8d1c30bc3f5b',
+      })
+    ).at(0);
+
+    const message = this.store.createRecord('bericht', {
+      verzonden: now,
+      aangekomen: now,
+      typeCommunicatie: type,
+      van: abb,
+      // auteur: ?, This should be someone from ABB, is this required?
```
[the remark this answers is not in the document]
Windvis: I removed it, it seems that `null` is a common value for it.

app/templates/berichtencentrum/berichten/new.hbs
Windvis: Done.

PR #344, opened 2023-11-14
Fix Berichtencentrum trigger VDDS

benjay10:
We need to make sure that the Bericht is saved after the Conversation and that the creator property is only stored *after* all other data. This can sometimes only be done by setting the creator to some placeholder at first.

### DIFF
--- a/app/controllers/berichtencentrum/berichten/new.js
+++ b/app/controllers/berichtencentrum/berichten/new.js
@@ -53,6 +53,8 @@ export default class BerichtencentrumBerichtenNewController extends Controller {
       inhoud: '',
       bijlagen: files,
     });
+    const creatorDefault = message.creator;
+    message.creator = 'pending';
     await message.save();
 
     const conversation = this.store.createRecord('conversatie', {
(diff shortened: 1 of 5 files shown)


app/controllers/berichtencentrum/berichten/new.js
```diff
@@ -53,6 +53,8 @@ export default class BerichtencentrumBerichtenNewController extends Controller {
       inhoud: '',
       bijlagen: files,
     });
+    const creatorDefault = message.creator;
+    message.creator = 'pending';
```
cecemel, line 57: nitpick: just `null` it?
less nitpick: please add a comment why you are doing so

cecemel (commented):
just a minor and less minor nitpick

app/controllers/berichtencentrum/berichten/new.js
cecemel: and that order matters (unfortunately)

benjay10: A comment I can do, but I also got complaints from other people that comments are annoying when there is too many for simple things.

About the `null`-ing: then the default value is being used automatically, which is exactly what I'm trying to prevent here.

PR #350, opened 2023-12-04
Enhance: loading state on send buttons

benjay10:
This PR converts the message sending actions to Ember Concurrency Tasks to make use of their state to easily add a loading indicator on the buttons.


Windvis (commented):
Nice! 👍 

The only suggestion I have is to also set a loadingMessage, otherwise it will show "Aan het laden" by default AFAIK which might look strange here?

PR #353, opened 2023-12-29
Add verenigingen app card and its external links

Riadabd:
## Ticket Number

DL-5538

## Ticket Description

A new app (`Verenigingen`) needs to be added to the module card in Loket for those who can access it.

## How to Test

TODO (Need to discuss some stuff first)

PR #355, opened 2024-01-09
Add contact app card and its external links

Riadabd:
## Ticket Number

DL-5537

## Ticket Description

Similar to https://github.com/lblod/frontend-loket/pull/353; a new app (`Contact- en Organisatiegegevens`) needs to be added to the module card in Loket for those who can access it.

## How to Test

We need to add a `sessionRole` for an organization(s) in order to trigger the external module card in the Loket homepage. Run the following migration inside app-digitaal-loket to grant an org the contact role:
```sparql
PREFIX ext:   <http://mu.semte.ch/vocabularies/ext/>
PREFIX foaf:  <http://xmlns.com/foaf/0.1/>

INSERT {
  GRAPH <http://mu.semte.ch/graphs/public> {
    ?account ext:sessionRole "abb_organisatieportaal_rol_3d" .
  }
}
WHERE {
  BIND(<BESTUURSEENHEID_URI> AS ?bestuurseenheid)

  ?persoon foaf:member ?bestuurseenheid ;
    foaf:account ?account .
}
```
Replace `BESTUURSEENHEID_URI` with a URI of your choice. On my local setup, I used:
```
http://data.lblod.info/id/bestuurseenheden/974816591f269bb7d74aa1720922651529f3d3b2a787f5c60b73e5a0384950a4
```
which corresponds to `Gemeente Aalst`.

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -18,6 +18,7 @@ const MODULE = {
   WORSHIP_DECISIONS_DB: 'LoketLB-databankEredienstenGebruiker',
   WORSHIP_ORGANISATIONS_DB: 'LoketLB-eredienstOrganisatiesGebruiker',
   VERENIGINGEN: 'abb_loketverenigingenapp',
+  CONTACT: 'abb_organisatieportaal_rol_3d'
 };
 
 export default class CurrentSessionService extends Service {
@@ -137,5 +138,9 @@ export default class CurrentSessionService extends Service {
 
   get canAccessVerenigingen() {
     return this.canAccess(MODULE.VERENIGINGEN);
+  }
+
+  get canAccessContact() {
+    return this.canAccess(MODULE.CONTACT);
   }
 }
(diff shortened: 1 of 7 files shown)


app/services/current-session.js
```diff
@@ -138,4 +139,8 @@ export default class CurrentSessionService extends Service {
   get canAccessVerenigingen() {
     return this.canAccess(MODULE.VERENIGINGEN);
   }
+
+  get canAccessContact() {
```
Windvis, line 143: I wonder if we should also verify if the actual url is set?

Similar to some of the other external apps: `!config.contactUrl.startsWith('{{')`

Riadabd: Yes, you're right; probably best to do so to avoid any issues. Should I add this snippet to `contact` only as part of this PR? Or can I add it to `verenegingen` and `public-services (LPDC)` as well?

Windvis: Yea, you can add it to the others as well if you want. For those it doesn't matter that much since they are already configured correctly (I assume).

Windvis (approved):

PR #359, opened 2024-01-30
Switch to Woodpecker CI

Windvis:
This replaces Drone with Woodpecker based on the templates from https://github.com/redpencilio/ci-templates


.woodpecker/release-controle.yml
```diff
@@ -0,0 +1,12 @@
+steps:
+  build:
+    image: woodpeckerci/plugin-docker-buildx
+    settings:
+      repo: ${CI_REPO}
+      tags: "${CI_COMMIT_TAG##v}"
```
Windvis, line 6: This needs the `-controle` suffix.

PR #364, opened 2024-02-27
[DL-5708] Fix an issue in the "create person" flow

Windvis:
We accidentally introduced a regression when resolving some [Appuniversum deprecations](https://github.com/lblod/frontend-loket/commit/cb16ed4a39b2a3b3ac2f5858b6d092aa683cf4ae#diff-1ddb3e656e7b094dd9777f1206d1292365678e3cac0c141b9e92c3b59dcda79bR184). The `setRijksregister` method was also used by the flow that sets initial data based on the provided rijksregisternummer, but this flow wasn't made compatible with the new code.

We now properly cover both scenarios.

### DIFF
--- a/app/components/shared/persoon/create-persoon.hbs
+++ b/app/components/shared/persoon/create-persoon.hbs
@@ -82,7 +82,7 @@
               mask="99.99.99-999.99"
               placeholder="_"
             )}}
-            {{on "input" this.setRijksregisternummer}}
+            {{on "input" this.handleRijksregisternummerChange}}
           />
           {{#if showError}}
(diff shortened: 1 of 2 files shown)


Riadabd:
Tested it locally, and the `create person` page shows up as expected now without any errors.

Windvis:
(Still testing a small tweak to this, got sidetracked by something 😄)

Windvis:
@Riadabd should be good. (I force pushed, so you need to do a hard reset to get the change, if you want to retest)

Riadabd:
Tested again and works as expected :+1:

PR #371, opened 2024-03-21
[DL-5790] Allow admin users to impersonate other accounts

Windvis:
This would allow us to put the controle environment behind an ACM/IDM login, while still allowing admins to see the data of certain organizations.

Requires https://github.com/lblod/app-digitaal-loket/pull/545

**TODOs**
- [x] Quick discussion of the implementation with the design team
- [x] determine what needs to happen when the user logs out with an active impersonation session -> Backend will clean up the whole session, so implicitly also the impersonated session.
- [x] Clean up the Git history

#### Flow 1
1. Set up this branch, and a local backend with the [corresponding backend PR](https://github.com/lblod/app-digitaal-loket/pull/545)
2. Start the app using the `start:controle` script: `npm run start:controle -- --proxy http://your-local-backend`
3. Log in as the mock admin user (should be the only option)
4. Notice it doesn't have access to any modules (we might want to improve the index page if that is actually a thing for the real admin users?)
5. click the "not-visible" icon in the header, next to the "admin" label
6. click "Simuleer een bestuurseenheid"
7. select a mock user to impersonate
8. on the index, click the eye icon, and it should display the name of the mock user
9. Click through the modules and verify that the data looks as expected (it should look the same as if logged in as that user)
10. Click the eye icon and click "Simuleer een andere bestuurseenheid"
11. Choose a different mock user
12. Click through some modules again and verify the data changed (and is the same as if logged in as that user directly)
13. Stop the impersonation by clicking the "Simulatie stoppen" link in the eye icon dropdown

#### Flow 2
1. Set up this branch, and a local backend with the [corresponding backend PR](https://github.com/lblod/app-digitaal-loket/pull/545)
2. Start the app using the normal start script: `npm start -- --proxy http://your-local-backend`
3. Log in as a non-admin mock user
4. Verify that the admin label and eye menu aren't visible
5. Try to access the impersonate route by entering the url directly '/impersonate'
6. Verify that the page can't be accessed
8. logout, and log in again as the mock admin user (search for "agentschap") and repeat the steps (the feature is only available in the "controle" build).

### DIFF
--- a/app/controllers/impersonate.js
+++ b/app/controllers/impersonate.js
@@ -1,0 +1,54 @@
+import Controller from '@ember/controller';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { restartableTask, task, timeout } from 'ember-concurrency';
+
+export default class ImpersonateController extends Controller {
+  @service impersonation;
+  @service router;
+  @service store;
+
+  queryParams = ['gemeente', 'page'];
+  @tracked model;
+  @tracked gemeente = '';
+  @tracked page = 0;
+  size = 10;
+
+  setup() {
+    this.queryStore.perform();
+  }
+
+  queryStore = task(async () => {
+    const filter = { provider: 'https://github.com/lblod/mock-login-service' };
+    if (this.gemeente) {
+      filter.gebruiker = { bestuurseenheden: this.gemeente };
+    }
+    const accounts = await this.store.query('account', {
+      include: 'gebruiker,gebruiker.bestuurseenheden',
+      filter: filter,
+      page: { size: this.size, number: this.page },
+      sort: 'gebruiker.achternaam',
+    });
+
+    this.model = accounts;
+  });
+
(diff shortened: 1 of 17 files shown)


app/controllers/impersonate.js
```diff
@@ -0,0 +1,54 @@
...
+    const filter = { provider: 'https://github.com/lblod/mock-login-service' };
+    if (this.gemeente) {
+      filter.gebruiker = { bestuurseenheden: this.gemeente };
+    }
+    const accounts = await this.store.query('account', {
+      include: 'gebruiker,gebruiker.bestuurseenheden',
+      filter: filter,
+      page: { size: this.size, number: this.page },
+      sort: 'gebruiker.achternaam',
+    });
```
Windvis, line 31: This is the same code that we use in the mock login route, so we can probably extract that into a function or service.

Riadabd (approved):
Looks good to go!

Tested both flows, and all expected behavior is working as intended.

PR #375, opened 2024-04-03
Remove controle image builds

Windvis:
Now that we have a dedicated admin user we no longer need a separate build.

Builds on #371


Riadabd (approved):
This one looks to go as well!

Tested using the same workflow as in https://github.com/lblod/frontend-loket/pull/371 (without the `controle` flow part), and all works as intended :+1:

PR #380, opened 2024-04-15
Hotfix/bump esff: to counter issue broken in prod

cecemel:
I looks like v0.92.2 is currently broken in loket. 
We suspect an issue with impersonation service.
The only other change is related to ESFF so we include the change in a hotfix build
https://github.com/lblod/frontend-loket/compare/v0.91.1...v0.92.2


Windvis:
I'll just add the entry to the changelog. I don't think this PR is needed anymore.

PR #381, opened 2024-04-15
Fix an issue when logging in with ACM/IDM

Windvis:
It seems that the ACM/IDM users don't have a relationship to their administrative units like the mock users do. This causes issues on QA and Prod environments. As a workaround we revert to the explicit API call that we did before. Ideally this would be resolved so the data is consistent between mock and real users, but this is a quick fix for the problem.


Windvis:
Ok, that wasn't the only issue.. Will create a feature image so it's easier to test without having to setup the ACM/IDM setup locally.

PR #382, opened 2024-04-15
Fix some issues after loging in as an ACM/IDM user

Windvis:
Follow-up to #381 since that didn't completely fix the problem.

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -40,7 +40,7 @@ export default class ApplicationController extends Controller {
       groupInfo += classification.label;
     }
 
-    if (group.naam) {
+    if (group?.naam) {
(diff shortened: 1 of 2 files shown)


Windvis:
These last fixes seem to have done it. Logging in works and doesn't throw exceptions.

app/controllers/application.js
```diff
@@ -40,7 +40,7 @@ export default class ApplicationController extends Controller {
       groupInfo += classification.label;
     }
 
-    if (group.naam) {
+    if (group?.naam) {
```
Windvis, line 43: Because we now do an extra API call, there is a small chance that the header is rerendered when the group isn't loaded yet (but the user is, so it tries to render the name).

PR #384, opened 2024-04-22
[DL-5849] Refactor the impersonation feature

Windvis:
The backend service changed implementations to work around some issues which requires us to update the logic in the frontend as well.

The backend now impersonates "transparently" which means the `/current` endpoints of the login services now return the impersonated account instead of the admin one. We then need to fetch the original admin account if an impersonation is active.

Needs the following branches to work:
- https://github.com/lblod/app-digitaal-loket/pull/561
- https://github.com/lblod/impersonation-service/pull/6

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -29,58 +29,31 @@ export default class CurrentSessionService extends Service {
   @service store;
   @service impersonation;
 
-  @tracked _account;
-  @tracked _user;
-  @tracked _group;
-  @tracked _groupClassification;
+  @tracked account;
+  @tracked user;
+  @tracked group;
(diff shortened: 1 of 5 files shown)


app/services/current-session.js
```diff
@@ -195,7 +184,11 @@ export default class CurrentSessionService extends Service {
   }
 
   get isAdmin() {
-    return this._roles.includes(ADMIN_ROLE);
+    let roles = this._roles;
+    if (this.impersonation.isImpersonating) {
+      roles = this.impersonation.originalRoles || [];
```
Windvis, line 189: I think the empty array fallback shouldn't be needed, not sure why I added that.

PR #390, opened 2024-05-08
Resolve Ember v4 deprecations

Windvis:
We should prepare the app for Ember v5 by resolving any deprecations we come across.

Update; some modules are moving to external apps, so we should probably wait until that is finished?


Windvis:
Rebase this once #396 is merged

PR #391, opened 2024-05-30
[DL-5866] Toezicht: Add a `DecisionArticles` field and validator

Windvis:
This adds a new custom form field that can be used to refer to multiple worship decisions.

Depends on https://github.com/lblod/app-digitaal-loket/pull/575

## To test:

1. ~~Run both this and the backend PR~~ The backend PR now includes the feature image of this PR.
2. ~~Apply the following patch to the active form file (check ACTIVE_FORM_FILE in docker-compose.yml)~~ Should no longer be needed with the latest backend commits.

<details>
<summary>Patch</summary>

```diff
diff --git a/config/semantic-forms/20240507113456-forms.ttl b/config/semantic-forms/20240507113456-forms.ttl
index e90b4334..7ab5a0f2 100644
--- a/config/semantic-forms/20240507113456-forms.ttl
+++ b/config/semantic-forms/20240507113456-forms.ttl
@@ -1737,6 +1737,42 @@ fields:eb089ca1-2c5b-42a4-a781-f009c46ac583 a form:Field ;
     form:displayType displayTypes:conceptSchemeSelector ;
     sh:group fields:aDynamicPropertyGroup .
 
+##########################################################
+# Worship Decisions cross reference selector
+##########################################################
+fields:788b9a98-684e-4d20-ae22-0ae610e6b494 a form:Field ;
+    mu:uuid "788b9a98-684e-4d20-ae22-0ae610e6b494" ;
+    sh:name "Artikels" ;
+    sh:order 102 ;
+    sh:path eli:has_part ;
+    form:validations
+      [ a form:RequiredConstraint ;
+        form:grouping form:Bag ;
+        sh:path eli:is_about ;
+        sh:path eli:has_part ;
+        sh:resultMessage "Dit veld is verplicht."@nl
+      ] ;
+    form:displayType displayTypes:worshipDecisionsCrossReferenceSelector ;
+    sh:group fields:aDynamicPropertyGroup .
+
+
+fields:type a form:Field ;
+    sh:name "Hidden field" ;
+    sh:order 102 ;
+    sh:path (<http://data.europa.eu/eli/ontology#has_part> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>);
+    sh:group fields:aDynamicPropertyGroup .
+
+fields:type_doc a form:Field ;
+    sh:name "Hidden field" ;
+    sh:order 1001 ;
+    sh:path (<http://data.europa.eu/eli/ontology#has_part> <http://data.europa.eu/eli/ontology#type_document>);
+    sh:group fields:aDynamicPropertyGroup .
+
+fields:refers_to a form:Field ;
+    sh:name "Hidden field" ;
+    sh:order 1001 ;
+    sh:path (<http://data.europa.eu/eli/ontology#has_part> <http://data.europa.eu/eli/ontology#refers_to>);
+    sh:group fields:aDynamicPropertyGroup .
 
 
 ##########################################################
@@ -2005,6 +2041,9 @@ fieldGroups:91045004-2066-40f3-aac1-bf2309eb2b99 a form:FieldGroup ;
                       
                       ###Eredienst selector###
                       fields:00a7bce6-2925-4794-93cb-ebd571b28831,
+
+                      ###Worship Decisions Cross Reference selector
+                      fields:788b9a98-684e-4d20-ae22-0ae610e6b494,
               
                       ###Rapportjaar###
                       fields:41737f90-02d6-4036-8d60-5d5b6ccf939c,
@@ -2655,6 +2694,9 @@ fieldGroups:09fd371f-2afe-4b14-81a9-e780876de077 a form:FieldGroup ;
                       ###Betreffend-bestuur-van-de-eredienst-(eredienst-selector)###
                       fields:00a7bce6-2925-4794-93cb-ebd571b28831,
 
+                      ###Worship Decisions Cross Reference selector
+                      fields:788b9a98-684e-4d20-ae22-0ae610e6b494,
+
                       ###Welk-beslissingsorgaan-nam-het-besluit?###
                       fields:4c7820f0-4011-4ab4-a16a-e128800e11bc,
 
@@ -2774,6 +2816,9 @@ fieldGroups:ae168679-82cf-4df6-8803-cbcbc6610f47 a form:FieldGroup ;
                       ###Eredienst selector###
                       fields:00a7bce6-2925-4794-93cb-ebd571b28831,
 
+                      ###Worship Decisions Cross Reference selector
+                      fields:788b9a98-684e-4d20-ae22-0ae610e6b494,
+
                       ###Gaat-het-over-het-origineel-document-of-over-een-wijziging?###
                       fields:a8f6a6cb-dbb8-488c-878d-05603791a9eb,
 
@@ -6255,6 +6300,9 @@ fieldGroups:abb02ef6-581d-4876-aa45-acd1fbc4f49b a form:FieldGroup ;
                       ###Eredienst selector###
                       fields:eb089ca1-2c5b-42a4-a781-f009c46ac583,
 
+                      ###Worship Decisions Cross Reference selector
+                      fields:788b9a98-684e-4d20-ae22-0ae610e6b494,
+
                       ###Datum-zitting/besluit###
                       fields:3dd6ed93-40f7-4d70-a6cb-f4de53dc8bfb,
 
@@ -6295,6 +6343,9 @@ fieldGroups:253be7ce-f723-4e82-8b24-3e55e0602aaa a form:FieldGroup ;
                       ###Eredienst selector###
                       fields:eb089ca1-2c5b-42a4-a781-f009c46ac583,
 
+                      ###Worship Decisions Cross Reference selector
+                      fields:788b9a98-684e-4d20-ae22-0ae610e6b494,
+
                       ###Datum-zitting/besluit###
                       fields:3dd6ed93-40f7-4d70-a6cb-f4de53dc8bfb,

```
</details>

3. Follow the test setup instructions in the backend PR: https://github.com/lblod/app-digitaal-loket/pull/575
4. Create a new decision as Gemeente Aalst and test the component.

### DIFF
--- a/app/components/supervision/decision-articles-field.gjs
+++ b/app/components/supervision/decision-articles-field.gjs
@@ -59,7 +59,7 @@ class DecisionArticlesField extends Component {
   constructor() {
     super(...arguments);
 
-    this.loadArticles();
+    this.loadArticles.perform();
   }
 
   get isReadOnly() {
@@ -99,7 +99,11 @@ class DecisionArticlesField extends Component {
     );
   }
 
-  loadArticles = () => {
+  loadArticles = task(async () => {
+    // This is needed to the rest of the method happens _after_ the willDestroy method of previous component instances,
+    // otherwise we would simply load the old data again before it was removed from the store.
+    await Promise.resolve();
+
     const {
       formStore,
       sourceNode,
@@ -147,7 +151,7 @@ class DecisionArticlesField extends Component {
     }
 
     this.articles = articles;
-  };
+  });
 
   createArticle = () => {
     const article = articleNode();
--- a/app/components/supervision/decision-articles-field.gjs
+++ b/app/components/supervision/decision-articles-field.gjs
@@ -18,8 +18,10 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { fn } from '@ember/helper';
 import { on } from '@ember/modifier';
+import { registerFormFields } from '@lblod/ember-submission-form-fields';
 import {
   RDF,
+  registerCustomValidation,
   SHACL,
   SKOS,
   validationsForFieldWithType,
@@ -38,7 +40,20 @@ const refersTo = ELI('refers_to');
 const prefLabel = SKOS('prefLabel');
 const shOrder = SHACL('order');
 
-export default class DecisionArticlesField extends Component {
+export function registerFormField() {
+  registerFormFields([
+    {
+      displayType:
+        'http://lblod.data.gift/display-types/worshipDecisionsCrossReferenceSelector',
+      edit: DecisionArticlesField,
+    },
+  ]);
+
+  registerFieldValidator();
+}
+
+/// Components
+class DecisionArticlesField extends Component {
   @tracked articles = [];
 
   constructor() {
@@ -68,18 +83,10 @@ export default class DecisionArticlesField extends Component {
   }
 
   get isDocumentTypeRequired() {
-    const shouldExcludeTypeLiteral = this.args.formStore.any(
+    return !isDocumentTypeOptional(
+      this.args.formStore,
       this.args.field.uri,
-      new NamedNode(
-        'http://lblod.data.gift/vocabularies/form-field-options/exclude-type_document',
-      ),
-      undefined,
       this.args.graphs.formGraph,
-    );
-
-    return (
-      typeof shouldExcludeTypeLiteral === 'undefined' ||
-      !Literal.toJS(shouldExcludeTypeLiteral)
     );
   }
 
@@ -115,7 +122,7 @@ export default class DecisionArticlesField extends Component {
       };
     });
 
-    // Add sh:order predicates if the sourceGraph doesn't contain them yet
+    // We add sh:order predicates if the sourceGraph doesn't contain them yet
     // (which could be possible if the decision wasn't created through the Toezicht module)
     if (
       !this.isReadOnly &&
@@ -194,7 +201,12 @@ export default class DecisionArticlesField extends Component {
         object: articleToRemove.node,
         graph: sourceGraph,
       },
-      ...formStore.match(articleToRemove.node, undefined, undefined, sourceGraph),
+      ...formStore.match(
+        articleToRemove.node,
+        undefined,
+        undefined,
+        sourceGraph,
+      ),
     ];
     formStore.removeStatements(triples);
   };
@@ -768,6 +780,26 @@ const RequiredPill = <template>
   <AuPill>Verplicht</AuPill>
 </template>;
 
+/// Template helpers
+function arrayIncludes(array, item) {
+  return array.includes(item);
+}
+
+function plusOne(number) {
+  return number + 1;
+}
+
+// TODO: Remove this once Appuniversum ships helpers for this
+// Source: https://github.com/appuniversum/ember-appuniversum/blob/f5bcb51c76333c4ac11858bdc17916f50f628bf5/addon/utils/date.ts#L1C1-L6C2
+function formatDate(date) {
+  const day = `${date.getDate()}`.padStart(2, '0');
+  const month = `${date.getMonth() + 1}`.padStart(2, '0');
+
+  return `${day}-${month}-${date.getFullYear()}`;
+}
+
+/// Utils
+
 // Source: https://github.com/lblod/ember-submission-form-fields/blob/5eb12a7794a70a04dfd1e8d392f0cc079d6aad72/addon/components/rdf-input-fields/concept-scheme-selector.js#L14C1-L18C2
 function byLabel(a, b) {
   const textA = a.label.toUpperCase();
@@ -777,10 +809,6 @@ function byLabel(a, b) {
 
 function byOrder(a, b) {
   return a?.order - b?.order;
-}
-
-function plusOne(number) {
-  return number + 1;
 }
 
 function articleNode() {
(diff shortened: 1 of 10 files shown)


app/styles/project/_c-article-details.scss
```diff
@@ -0,0 +1,5 @@
+.article-details {
```
Windvis, line 1: This should probably be renamed to be more specific to the field. Will do when the final name of the component is decided.

app/templates/supervision/submissions/edit.hbs
```diff
@@ -46,16 +46,15 @@
...
-      <div class="au-o-grid__item au-u-1-2@medium au-u-1-1">
-        <SubmissionForm
-          @form={{this.form}}
-          @show={{this.model.submitted}}
-          @graphs={{this.graphs}}
-          @sourceNode={{this.sourceNode}}
-          @formStore={{this.formStore}}
-          @forceShowErrors={{this.forceShowErrors}} />
-      </div>
+    <div class="au-u-max-width-small">
```
Windvis, line 49: I replaced the "50% of the page" width with a max-width instead since otherwise the table would show scrollbars, even on screens that are wide enough if the container had more space.

app/components/supervision/worship-decisions-cross-reference-selector.gjs
```diff
@@ -0,0 +1,742 @@
...
+          @isReadOnly={{@isReadOnly}}
+          @error={{if this.hasTypeError "Dit veld is verplicht"}}
+          @required={{true}}
+          class="au-u-margin-bottom"
+        >
+          <:label>Artikeltype</:label>
+        </ConceptSchemeSelect>
+
+        <AuTable @size="small" class="au-table-test">
+          <:title>Besluiten</:title>
```
Windvis, line 342: Not sure if "Besluiten" is the correct term here. The article type concept scheme refers to "Documenten" but technically these are decisions so that's what I went for for now.

Not sure which term to use.

app/routes/supervision/submissions/edit.js
```diff
@@ -2,15 +2,31 @@ import Route from '@ember/routing/route';
...
   @service store;
 
+  constructor() {
+    super(...arguments);
+
+    registerFormFields([
+      {
+        displayType:
+          'http://lblod.data.gift/display-types/worshipDecisionsCrossReferenceSelector',
+        edit: WorshipDecisionsCrossReferenceSelector,
```
Windvis, line 25: The component can probably use a better name. Depends a bit on the "document vs decision" naming.

app/components/supervision/worship-decisions-cross-reference-selector.gjs
```diff
@@ -0,0 +1,742 @@
...
+        object: BESLUIT('Artikel'),
+        graph: sourceGraph,
+      },
+      {
+        subject: sourceNode,
+        predicate: hasPart,
+        object: article,
+        graph: sourceGraph,
+      },
+      // TODO: should we add "order" predicates?
```
Windvis, line 108: We should, but not sure which predicate to use. `qb:order` vs `sh:order`? I think we use both in our apps.

app/components/supervision/worship-decisions-cross-reference-selector.gjs
```diff
@@ -0,0 +1,742 @@
...
+  return textA < textB ? -1 : textA > textB ? 1 : 0;
+}
+
+function plusOne(number) {
+  return number + 1;
+}
+
+function articleNode() {
+  // TODO, double check if this base url is fine
+  return new NamedNode(`http://data.lblod.info/besluit-artikel/${uuid()}`);
```
Windvis, line 674: @cecemel Not sure what the policy is for this. I saw we used data.lblod.info in a similar case, but not sure about the path. Let me know if this needs to be changed.

app/components/supervision/worship-decisions-cross-reference-selector.gjs
Windvis: Went with `sh:order` since apparently it doesn't really matter.

app/components/supervision/worship-decisions-cross-reference-selector.gjs
cecemel: i think we solved this

app/routes/supervision/submissions/edit.js
cecemel: Close enough

app/components/supervision/worship-decisions-cross-reference-selector.gjs
```diff
@@ -0,0 +1,810 @@
...
+      <AuModal @modalOpen={{true}} @closeModal={{@onClose}} @overflow={{true}}>
+        <:title>Besluiten zoeken</:title>
+        <:body>
+          <ConceptSchemeSelect
+            @formStore={{@formStore}}
+            @metaGraph={{@metaGraph}}
+            {{! // TODO: there are 2 different "betreffende bestuur van de eredienst" fields in the form and both have a different concept scheme, so this might need to be adjustable if this is correct }}
+            {{! bestuurseenheden }}
+            @conceptScheme="http://lblod.data.gift/concept-schemes/2e136902-f709-4bf7-a54a-9fc820cf9f07"
+            {{!-- @conceptScheme="http://lblod.data.gift/concept-schemes/164a27d5-cf7e-43ea-996b-21645c02a920" {{! bestuurseenheden }} --}}
```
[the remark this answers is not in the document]
Windvis: The current version is correct, so we can remove the commented code.

app/components/supervision/decision-articles-field.gjs
```diff
@@ -0,0 +1,939 @@
...
+      }
+
+      const type = store.any(articleNode, documentType, undefined, sourceGraph);
+
+      return Boolean(type);
+    })
+    .every(Boolean);
+
+  return areArticlesValid;
+}
```
Windvis, line 939: Ideally this validator also runs in the backend, but the validation service still runs an [older version](https://github.com/lblod/validate-submission-service/blob/f429c330d3c5c73533d2cd141e9a55a2587e06f8/package.json#L13) of the submission-form-helpers package, so it needs some updates first.

Windvis:
Hmm, even with the 2 display type setup I'm running into state persistency issues.

To reproduce:
- select "Schorsing" in the "type dossier" field
- add an article and add some documents to it
- change the "type dossier" to "Besluit over Budget(wijziging) eredienstbestuur"
- notice that the article and documents are still there, even though they should have been cleared

It seems that the `willDestroy` of the original component instance runs _after_ the constructor of the new one, so the loadData method simply finds the data again before it is removed from the store. Not sure if this an Ember quirk, or something in our field handling logic that causes this.

PR #393, opened 2024-06-24
Fix: subsidies always visible

wolfderechter:
## ID
DGS-316

## Description
Since subsidies will move to subsidiepunt as an external app instead of sub-app, we need to always show the subsidies tab

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -4,7 +4,6 @@ import { setContext, setUser } from '@sentry/ember';
 import config from 'frontend-loket/config/environment';
 import { loadAccountData } from 'frontend-loket/utils/account';
 import { SHOULD_ENABLE_SENTRY } from 'frontend-loket/utils/sentry';
-import isFeatureEnabled from 'frontend-loket/helpers/is-feature-enabled';
 
 const MODULE_ROLE = {
   SUPERVISION: 'LoketLB-toezichtGebruiker',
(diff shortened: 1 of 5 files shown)


app/services/current-session.js
```diff
@@ -151,17 +150,6 @@ export default class CurrentSessionService extends Service {
     return this.canAccess(MODULE_ROLE.PERSONEELSBEHEER);
   }
 
-  get canAccessSubsidies() {
-    if (isFeatureEnabled('subsidies-external')) {
```
cecemel, line 155: can you keep the toggle logic here please (and also in the templates)? 
- so we keep it consistent
-  also; this will allow us to choose when to deploy loket. This change implies we NEED to deploy loket and subsidiepunt at the same time. I don't want that. I want to first deploy loket; then at own pace do the migration then do the cutover

cecemel (changes requested):

Windvis:
If it's a standalone thing it probably makes sense to always hide info about it on the login page?

https://github.com/lblod/frontend-loket/blob/2e054f6892887f607330fd434561a423e82bc43e/app/templates/login.hbs#L300-L330

I think that part needs to be updated to either hide it if the feature is active, or display the `SubsidiePunt` name instead, if we still want to keep the card.

wolfderechter:
> If it's a standalone thing it probably makes sense to always hide info about it on the login page?
> 
> https://github.com/lblod/frontend-loket/blob/2e054f6892887f607330fd434561a423e82bc43e/app/templates/login.hbs#L300-L330
> 
> I think that part needs to be updated to either hide it if the feature is active, or display the `SubsidiePunt` name instead, if we still want to keep the card.

Marie definitely wants to keep it there, so I have updated it to say 'SubsidiePunt' when needed

Windvis (approved):

PR #394, opened 2024-07-02
[DL-6042] Add cards to the login page for all external apps

Windvis:
It seems users get confused when they don't see the cards listed on the login page so we add them all, even though it's technically not part of loket itself.

The following cards have been added:
- Databank Erediensten
- Organisaties Erediensten
- Verenigingen
- Contact- en Organisatiegegevens
- Open Proces Huis

### DIFF
--- a/app/templates/login.hbs
+++ b/app/templates/login.hbs
@@ -187,8 +187,9 @@
                   </AuHeading>
                 </c.header>
                 <c.content>
-                  <p class="au-u-margin-top-tiny">Hou de mandaten binnen de
-                    erediensten bij.</p>
+                  <p class="au-u-margin-top-tiny">
+                    Hou de bedienaren binnen de erediensten bij.
+                  </p>
                 </c.content>
                 <c.footer>
                   <p>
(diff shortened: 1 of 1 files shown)


app/templates/login.hbs
```diff
@@ -187,8 +187,9 @@
                   </AuHeading>
                 </c.header>
                 <c.content>
-                  <p class="au-u-margin-top-tiny">Hou de mandaten binnen de
-                    erediensten bij.</p>
+                  <p class="au-u-margin-top-tiny">
+                    Hou de bedienaren binnen de erediensten bij.
```
Windvis, line 191: mandaten -> bedienaren since this is about "Bedienarenbeheer"

app/templates/login.hbs
```diff
@@ -365,6 +367,165 @@
...
+                    Consulteer uw verenigingen, zoals gekend in het
+                    Verenigingsregister, en beheer uw erkenningen.
+                  </p>
+                </c.content>
+                <c.footer>
+                  <p>
+                    <AuLinkExternal
+                      @skin="button"
+                      @icon="link"
+                      href="https://www.vlaanderen.be/lokaal-bestuur/nieuwsberichten/binnenkort-nieuwe-module-verenigingen-in-loket-voor-lokale-besturen"
```
Windvis, line 458: Not sure if we should be using "nieuwsberichten" but I couldn't find an actual page about this card. Alternatively we could hide the link, but I think the news article explains it pretty well already.

Windvis: Discussed this with Pieter. We should use the general loket link instead: https://www.vlaanderen.be/lokaal-bestuur/organisatie-en-werking/loket-voor-lokale-besturen (for both "verenigingen" and "contact" modules).

PR #397, opened 2024-07-04
Also add the `forDecisionType` param to the `/related-document-information` call

Windvis:
This makes it easier for the backend to do its thing.


cecemel (approved):
thanks

PR #399, opened 2024-07-11
Add a status filter to the "Toezicht" overview page

Windvis:
This makes it (a little) easier to find the submission we want.

image

Builds on #398


Windvis:
(Moving it back to draft since we might want to add more filters? Needs to be discussed further)

benjay10:
As already mentioned by Riad, the menu indeed does not close after selecting an option. You can only select a single option, so the menu can be closed immediately after selecting one.

Windvis:
> As already mentioned by Riad, the menu indeed does not close after selecting an option. You can only select a single option, so the menu can be closed immediately after selecting one.

Thanks, should be good now :+1:

benjay10:
Yes, seems fixed now.

Riadabd (approved):
Looks good now :+1:

PR #400, opened 2024-07-11
Improvements to buttons, links and icons

Windvis:
Loket is an old project which predates some of the features in Appuniversum. This changes some of the old patterns to the new ones which improves consistency, styling and alignment.

This:
- removes au-c-link with `AuLink` or `AuLinkExternal` where possible
- replaces au-c-button with `AuButton` where possible
- replaces `AuIcon` with `@icon` where possible
- adds icons where needed to make things more consistent between modules

### DIFF
--- a/app/components/mandatenbeheer/mandataris-table.hbs
+++ b/app/components/mandatenbeheer/mandataris-table.hbs
@@ -73,7 +73,15 @@
           {{/if}}
         </td>
       {{/if}}
-      <td><LinkTo @route={{@editRoute}} @model={{row.isBestuurlijkeAliasVan.id}} class="au-c-link"> Bewerk </LinkTo></td>
+      <td>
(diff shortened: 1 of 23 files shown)


app/components/mandatenbeheer/mandataris-table.hbs
```diff
@@ -73,7 +73,15 @@
           {{/if}}
         </td>
       {{/if}}
-      <td><LinkTo @route={{@editRoute}} @model={{row.isBestuurlijkeAliasVan.id}} class="au-c-link"> Bewerk </LinkTo></td>
+      <td>
+        {{#if row.isBestuurlijkeAliasVan.id}}
```
Windvis, line 77: This extra check is needed because it's an async relationship which is loaded by the template, and for some reason an error is thrown if null is passed into the `AuLink` component as the model value, but it works under LinkTo. :thinking:

PR #405, opened 2024-08-03
[DL-4540, DL-4069] Fix sorting in various places

benjay10:
**[DL-4540]** and **[DL-4069]**

The tables in the "Mandatenbeheer", "Bedienarenbeheer", "Toezicht" and "Berichtencentrum" modules all showed similar problems while sorting (alphabetically). The sorting would look good at first glance, but sometimes a single entry would be in the wrong place. Sometimes the sorting would look completely random.  
The root cause of this issue is a bug in Virtuoso, but can be solved by sorting on lowercase conversions of the fields.

**Note:** some edge case observed while testing that is not fixed: in the "Berichtencentrum" module, the "Betreft" column might still look improperly sorted. This could be caused by whitespaces at the beginning of the field that are not shown in the table because HTML collapses/trims spaces automatically. These spaces are in the database where the sorting happens, putting them at the top of the alphabetical list.

This PR also fixes another issue: sorting on "Type Dossier" of Submission in the module "Toezicht" caused an error. The sorting property did not exist in mu-cl-resources, because it is computed in the component code from a has-many relation. This has been changed to the correct model property, which seems to output useful results from mu-cl-resources. It is not clear whether this is a sound solution, because the property points to a has-many-relation.

### DIFF
--- a/app/components/berichtencentrum/conversatie-table.hbs
+++ b/app/components/berichtencentrum/conversatie-table.hbs
@@ -12,12 +12,12 @@
    as |t|>
   <t.content as |c|>
     <c.header>
-      <AuDataTableThSortable @label='Betreft' @field='betreft' @currentSorting={{@sort}} />
-      <AuDataTableThSortable @label='Type communicatie' @field='currentTypeCommunicatie' @currentSorting={{@sort}} />
-      <AuDataTableThSortable @label='Dossiernummer' @field='dossiernummer' @currentSorting={{@sort}} class="au-u-visible-small-up" />
+      <AuDataTableThSortable @label='Betreft' @field=':no-case:betreft' @currentSorting={{@sort}} />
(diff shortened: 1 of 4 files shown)


Windvis:
> This PR also fixes another issue: sorting on "Type Dossier" of Submission in the module "Toezicht" caused an error. The sorting property did not exist in mu-cl-resources, because it is computed in the component code from a has-many relation. This has been changed to the correct model property, which seems to output useful results from mu-cl-resources. It is not clear whether this is a sound solution, because the property points to a has-many-relation.

Strange, I wonder why it was implemented this way. We should probably remove the `decisionsType` relationship from the FormData class then? It seems strange that a `FormData` could have more than one type though :thinking: .

Anyways, I don't know the history of this code and it's out of scope of this PR. Good that you noticed :+1:

Windvis (approved):

PR #406, opened 2024-08-06
Update to Ember 5.8 LTS

Windvis:
This updates the app to the Ember v5.8 LTS blueprint which is the current LTS release. I did some basic testing of the app and things seem to work as expected, but it's hard to cover all flows manually, so maybe I missed something. The testing team should do a general test.

Supersedes #390 since that change is included here.


package.json
```diff
@@ -51,66 +52,68 @@
...
     "ember-concurrency": "^3.1.1",
-    "ember-data": "~4.12.3",
+    "ember-data": "~5.3.3",
     "ember-fetch": "^8.1.2",
+    "ember-inflector": "^4.0.2",
     "ember-load-initializers": "^2.1.2",
     "ember-modifier": "^4.1.0",
     "ember-moment": "^9.0.0",
-    "ember-mu-transform-helpers": "^2.1.2",
+    "ember-mu-transform-helpers": "^2.1.3",
```
Windvis, line 77: This was needed for the EmberData update, since the addon was using private import paths that were changed.

PR #407, opened 2024-08-08
Resolve ember data v5 deprecations

Windvis:
This:
- resolves some EmberData v5 deprecations
- modernizes and simplifies the custom inflector rule setup
- removes unused EmberData models


app/config/custom-inflector-rules.js
Windvis, line 1: We now assume a model uses an English name by default which simplifies the setup a lot. Most of the config was added since the models originally had Dutch names.

PR #408, opened 2024-08-27
feat: update mandatenbeheer tegels

wolfderechter:
## ID
DL-6046

## Description

Adds a feature flag to toggle the existing mandatenbeheer tile, as well as adding a new mandatenbeheer tile for the external mandatenbeheer application.

docker-compose.override.yml for PROD:
```
EMBER_MANDATENBEHEER_EXTERNAL_URL: "mandatenbeheer.lokaalbestuur.vlaanderen.be
```

docker-compose.override.yml for QA:
```
EMBER_MANDATENBEHEER_EXTERNAL_URL: "https://mandatenbeheer.lblod.info/"
```

docker-compose.override.yml for DEV:
```
EMBER_MANDATENBEHEER_EXTERNAL_URL: "https://dev.mandatenbeheer.lblod.info/"
```

---

Make sure to add the following environment variable for the existing mandatenbeheer tile:
```
EMBER_MANDATENBEHEER: "true"
```
To disable the tile, simply change the value to false.

Note: no redirect logic was implemented since the new mandatenbeheer application is too different, so wasn't worth the effort.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -134,7 +134,11 @@ export default class CurrentSessionService extends Service {
   }
 
   get canAccessMandaat() {
-    return this.canAccess(MODULE_ROLE.MANDATENBEHEER);
+    return this.canAccess(MODULE_ROLE.MANDATENBEHEER) && config.mandatenbeheer;
+  }
+
+  get canAccessMandaatExternal() {
+    return !config.mandatenbeheerExternalUrl.startsWith('{{');
   }
 
   get canAccessBerichten() {
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -134,7 +134,10 @@ export default class CurrentSessionService extends Service {
   }
 
   get canAccessMandaat() {
-    return this.canAccess(MODULE_ROLE.MANDATENBEHEER) && config.mandatenbeheer === 'true';
+    return (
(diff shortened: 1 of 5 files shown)


app/services/current-session.js
```diff
@@ -134,7 +134,11 @@ export default class CurrentSessionService extends Service {
   }
 
   get canAccessMandaat() {
-    return this.canAccess(MODULE_ROLE.MANDATENBEHEER);
+    return this.canAccess(MODULE_ROLE.MANDATENBEHEER) && config.mandatenbeheer;
```
Windvis, line 137: Isn't `config.mandatenbeheer` always truthy unless we set it to an empty string in the docker-compose.yml file?

I _think_ this will always return `true` with the current code, instead of only when it's set to `'true'`?

```suggestion
    return this.canAccess(MODULE_ROLE.MANDATENBEHEER) && config.mandatenbeheer === 'true';
```

Windvis:
This should be good to go once you run Prettier locally :+1:

PR #411, opened 2024-10-10
[DL-6207] Add a new DecisionDocumentsField component

Windvis:
This is a simplified variant of the `DecisionsArticlesField` component that only contains the document selection functionality.

Depends on:
- https://github.com/lblod/app-digitaal-loket/pull/605
- https://github.com/lblod/enrich-submission-service/pull/22
- https://github.com/lblod/worship-decisions-cross-reference-service/pull/3
- https://github.com/lblod/manage-submission-form-tooling/pull/54

image

### DIFF
--- a/app/components/supervision/fields/decision-documents-field.gjs
+++ b/app/components/supervision/fields/decision-documents-field.gjs
@@ -1,0 +1,276 @@
+import AuButton from '@appuniversum/ember-appuniversum/components/au-button';
+import AuHelpText from '@appuniversum/ember-appuniversum/components/au-help-text';
+import AuIcon from '@appuniversum/ember-appuniversum/components/au-icon';
+import AuLabel from '@appuniversum/ember-appuniversum/components/au-label';
+import AuLinkExternal from '@appuniversum/ember-appuniversum/components/au-link-external';
+import AuLoader from '@appuniversum/ember-appuniversum/components/au-loader';
+import AuTable from '@appuniversum/ember-appuniversum/components/au-table';
+import { AddIcon } from '@appuniversum/ember-appuniversum/components/icons/add';
+import { BinIcon } from '@appuniversum/ember-appuniversum/components/icons/bin';
+import { InfoCircleIcon } from '@appuniversum/ember-appuniversum/components/icons/info-circle';
+import Component from '@glimmer/component';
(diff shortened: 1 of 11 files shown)


app/components/supervision/fields/decision-documents-field.gjs
```diff
@@ -0,0 +1,287 @@
...
+import { triplesForPath } from '@lblod/submission-form-helpers';
+import { task } from 'ember-concurrency';
+import worshipDecisionsDatabaseUrl from 'frontend-loket/helpers/worship-decisions-database-url';
+import { DCT, RDF } from 'frontend-loket/rdf/namespaces';
+import { formatDate } from 'frontend-loket/utils/date';
+import { isRequiredField } from 'frontend-loket/utils/semantic-forms';
+import { AddDocumentsModal } from './-shared/add-documents-modal';
+import { extractDocumentsFromTtl } from './-shared/utils';
+
+const relation = DCT('relation');
```
Windvis, line 25: We can probably just use the path assigned to the field instead of hardcoding this. That would allow us to use it with a different predicate if needed.

app/components/supervision/fields/decision-documents-field.gjs
```diff
@@ -0,0 +1,287 @@
...
+    await Promise.resolve();
+
+    console.log(
+      triplesForPath({
+        store: this.args.formStore,
+        formGraph: this.args.graphs.formGraph,
+        sourceGraph: this.args.graphs.sourceGraph,
+        sourceNode: this.args.sourceNode,
+      }),
+    );
```
Windvis, line 88: I still need to see if I can use this instead of manually looking at the store. Delete otherwise.

app/components/supervision/fields/decision-documents-field.gjs
```diff
@@ -0,0 +1,287 @@
...
+      relation,
+      undefined,
+      sourceGraph,
+    );
+
+    const documentPromises = triples.map(async (triple) => {
+      const document = triple.object;
+
+      const url = new URL(
+        '/worship-decisions-cross-reference/document-information',
```
Windvis, line 107: This seems to work, but I'm not sure why because I didn't update that part of the backend yet :smile:

benjay10 (approved):

PR #412, opened 2024-10-17
[DL-6305] Update to the Ember v5.12 blueprints

Windvis:
This is the last release before v6, which is a good place to resolve all deprecations.

This also includes an updated linting setup since the updated deps in v5.12 caused some issues with the .gjs setup.


eslint.config.mjs
Windvis, line 1: This setup needs more work. I copied it from the ember-cli repo, but it's not 100% working as expected yet.

PR #414, opened 2024-11-12
Update to Appuniversum v3.7.0

Windvis:
This release fixes some v5.12 deprecations so we are one step closer to being v6.0 compatible.


Windvis:
Hmm, some dependency conflicts. We'll probably need to downgrade to the v1 addon version since ember-data-table doesn't depend on ember-auto-import v2 yet (so it can't use v2 addons)..

PR #416, opened 2024-11-18
[DL-6310] Improve the error handling in the toezicht module

Windvis:
At the moment we dont really do any error handling which makes it not very clear to the user that something went wrong while saving and submitting a form. We should display some error messages to make this more obvious.

I'm also experimenting with EmberData's new RequestManager instead of using vanilla fetch, since it already handles some error states for us.

**Test instructions**
- Saving, submitting and deleting submissions should still work as expected
- Testing the error states is tricky without modifying code to throw errors. You could use your dev tools to mimic the no-internet case before clicking any of the buttons.

### DIFF
--- a/app/controllers/supervision/submissions/edit.js
+++ b/app/controllers/supervision/submissions/edit.js
@@ -7,22 +7,23 @@ import {
   delGraphFor,
   addGraphFor,
 } from '@lblod/ember-submission-form-fields';
-import fetch from 'fetch';
 import { DELETED_STATUS } from '../../../models/submission-document-status';
 import { task } from 'ember-concurrency';
-import { inject as service } from '@ember/service';
+import { service } from '@ember/service';
+import fetchManager from 'frontend-loket/utils/fetch-manager';
+import fetch from 'frontend-loket/utils/fetch';
 
 export default class SupervisionSubmissionsEditController extends Controller {
   @service currentSession;
   @service store;
-  @service() router;
+  @service router;
+  @service toaster;
 
   @tracked datasetTriples = [];
   @tracked addedTriples = [];
   @tracked removedTriples = [];
   @tracked forceShowErrors = false;
   @tracked isValidForm = true;
-  @tracked recentlySaved = false;
 
   constructor() {
     super(...arguments);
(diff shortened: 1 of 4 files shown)


app/templates/supervision/submissions/edit.hbs
```diff
@@ -20,26 +20,6 @@
...
-      {{#if this.delete.isRunning}}
-        <li class="au-o-grid__item au-u-1-2 au-u-2-6@medium">
-          <AuAlert @icon="info-circle" @title="Aan het verwijderen..."  @skin="info" @size="small"></AuAlert>
-        </li>
-      {{/if}}
-      {{#if this.recentlySaved}}
-        <li class="au-o-grid__item au-u-1-2 au-u-1-6@medium">
-          <AuAlert @icon="info-circle" @title="Opgeslagen."  @skin="success" @size="small"></AuAlert>
-        </li>
-      {{/if}}
```
Windvis, line 42: These are no longer needed in modern Appuniversum, since they can be replaced with toasters and button loading messages.

app/utils/fetch.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * A simple fetch wrapper that throws an error if the response is not "ok"
+ */
+export default async function fetch() {
+  const response = await globalThis.fetch(...arguments);
+
+  if (!response.ok) {
```
Windvis, line 7: This handles the most common case where `window.fetch` considers 400 - 500 errors "ok" and only throws due to connectivity issues.

Windvis:
CI faillures are due to the lock file changes #427 I'll rebase once that's resolved.

app/controllers/supervision/submissions/edit.js
```diff
@@ -109,43 +110,57 @@ export default class SupervisionSubmissionsEditController extends Controller {
...
+      // TODO: use the request manager once that supports non-json responses
+      yield fetch(`/submissions/${this.model.submission.id}`, {
+        method: 'DELETE',
+      });
+      this.toaster.success(undefined, 'Dossier verwijderd', {
+        timeOut: 3000,
+      });
+      this.router.transitionTo('supervision.submissions');
+    } catch {
+      this.toaster.error(
```
Windvis, line 138: At the moment we simply catch the exception and show a message. 

It might be nice for some cases to still log the exception to Sentry, in case it's a server error? I'm not sure since the server logs should also contain the errors already. There are also types of issues which we don't want to log, for example network connection problems, if those aren't things we can fix.

I consider this out of scope for the current PR though, we simply want a visual error state so users know something went wrong.

claire-lovisa (approved):
I've been able to test the error states locally by turning off virtuoso just before saving/submitting and everything was working as advertised! Happy to have discovered the toasters that I somehow missed before!

PR #421, opened 2024-12-03
Chose concept scheme depending on decision type

claire-lovisa:
For extra complexity, the concept scheme we need to get the EB and/or CB changes depending on the decision type.

If the decision is a `Opvragen bijkomende inlichtingen eredienstbesturen (met als gevolg stuiting termijn)` or a `Schorsing beslissing eredienstbesturen`, the selector needs to show EB and CB linked to the admin unit. But if the decision of any other type, we need to only show EB as the CB are "bundling" EB's submissions.

I am not sure if this is the best place to do this. But I also feel like updating the enrich-submission-service with an extra concept scheme that depends on the decision type [around here](https://github.com/lblod/enrich-submission-service/blob/ab0500e74582c155aaef2c17ea3d3b3bb0a64705/lib/enricher.js#L47) isn't perfect... Any opinion before I move further? :sweat_smile:


app/components/supervision/fields/decision-articles-field.gjs
```diff
@@ -98,6 +98,21 @@ class DecisionArticlesField extends Component {
     );
   }
 
+  get eenheidConceptScheme() {
```
Windvis, line 101: 1 small nitpick; I would move this gettor to the ArticleDetails component since it's only needed there and it already receives the decisionType as well.

claire-lovisa: You're right!

Windvis (approved):
Code looks good to me :+1: I haven't tested it though, but I trust that the correct data is contained in those codelists now :smile:

cecemel (approved):

PR #426, opened 2025-02-14
Berichtencentrum - prevent empty or too small files from being uploaded

aatauil:
**ID**
DL-6240

**Description**
This bump ember-appuniversum and sets a min of 1kb for file upload

**How to setup**
Setup like usual

**How to test**
login with any bestuurseenheid that has access to berichtencentrum, open berichtcentrum, choose a bericht, multiple ways for the file upload zone to appear, try uploading an empty file (0bytes) try uploading a file under 1kb but not 0

PR #427, opened 2025-02-18
Update the lock file

Windvis:
Not sure what's up here. When I run npm install on node 20 I get these lock changes.

Maybe I forgot to commit some changes when updating the blueprint a while ago?


Windvis:
Probably not needed once #426 is merged, since that also includes lock changes.

PR #428, opened 2025-03-05
[DL-3685] Address selection improvements

Windvis:
The app uses an address selector component in several modules. This code predates the creation of the [`@lblod/ember-address-search` addon](https://github.com/lblod/ember-address-search), so we can now remove a lot of it by switching to the addon version.

While testing I also noticed a bug in the eredienst mandatenbeheer version, so I fixed that as well.

### Test instructions
There are 3 modules that use the address selector, so there are different testing flows
#### Leidinggevendenbeheer
1. Log in as a gemeente
2. Go to the leidinggevendenbeheer module
3. Click the button next to one of the functies
4. Click the "bewerk" button next to the contact information at the top
5. Add and edit the address and see if saving and loading still works as expected
6. Also try this with an address that has a "busnumber" (I used "Nijlse baan 3A", but feel free to use something)

#### Eredienst mandatenbeheer / Eredienst bedienarenbeheer
These are separate modules, but the address selection flow is very similar.

1. Log in as a worship service
2. Go to mandatenbeheer or bedienarenbeheer module
3. Select a bedienaar / mandaat or create a new one if there aren't any yet
4. edit the contact info and add / edit the address


benjay10 (approved):
This worked perfectly on my machine

Riadabd (approved):
Works as expected on my end as well :+1:

Windvis:
Thanks for the reviews!

PR #433, opened 2025-06-12
[DL-6646] Allow users to update the modified date without any table changes

Windvis:
Some organizations don't change much every year so they simply want to mark the data as correct without having to do any changes in the tables.

We now display a button at the top that can be used to mark the data as up-to-date.

Demo video:
[Screencast from 2025-06-12 15-41-22.webm](https://github.com/user-attachments/assets/cbb30ada-f3be-4e12-9f3a-9b1f5e8c9bb3)


claire-lovisa (approved):
The changes work well, let's wait for Nele's confirmation on the behavior before merging

PR #461, opened 2026-04-09
[DL-7261] Update user manual links

Windvis:
This replaces all references to the old Gitbook with the new links. 

The "help" link now also points to the most specific one, based on where the user is located in the app.

### DIFF
--- a/app/helpers/relevant-user-manual-link.js
+++ b/app/helpers/relevant-user-manual-link.js
@@ -1,0 +1,33 @@
+import Helper from '@ember/component/helper';
+import { service } from '@ember/service';
+
+const defaultManualLink =
+  'https://abb-vlaanderen.gitbook.io/hoofdloket-handleiding-loket-lokale-besturen';
+
+/**
+ * A helper that looks the meta data of all active routes and retrieves the userManualLink value and falls back to the generic loket manual if none is found.
+ */
+export default class RelevantUserManualLink extends Helper {
+  @service router;
+
+  compute() {
+    let manualLink = defaultManualLink;
+    let route = this.router.currentRoute;
+
+    do {
+      if (route.metadata?.userManualLink) {
+        manualLink = route.metadata.userManualLink;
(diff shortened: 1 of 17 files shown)


app/templates/login.hbs
Windvis, line 1: Technically these links aren't visible anymore, since the new-loket login page is displayed on all environments, but it felt wrong to leave the old links in the code.

app/helpers/relevant-user-manual-link.js
```diff
@@ -0,0 +1,32 @@
+import Helper from '@ember/component/helper';
+import { service } from '@ember/service';
+
+const defaultManualLink = 'https://abb-vlaanderen.gitbook.io/handleiding-loket';
```
wolfderechter: shouldn't it be https://abb-vlaanderen.gitbook.io/hoofdloket-handleiding-loket-lokale-besturen/ ?

app/templates/supervision/submissions/index.hbs
```diff
@@ -28,6 +28,13 @@
   />
 </div>
 
+<AuAlert @skin="warning" @icon="alert-triangle" @size="small">
```
wolfderechter: should this be an info alert instead of warning?

app/helpers/relevant-user-manual-link.js
Windvis: Yes, it seems I copied the wrong one, thanks

app/templates/supervision/submissions/index.hbs
Windvis: I just matched the alert setup that was already present in some of the other modules since that's what they asked in the ticket, but I'm not really convinced by it myself.

Windvis: Ok, you were right, I changed both to an info and changed the icon:

image

Still don't really like it, but better than before, thanks.

PR #464, opened 2026-04-17
[DL-7300]: Use relevantAdministrativeUnit filter by default

NikiDeVisscher:


app/templates/search.hbs
```diff
@@ -7,13 +7,22 @@
...
+          />
+        </div>
         <div class="au-o-box au-o-box--small">
           <ConceptSchemeFilter
             @conceptScheme={{this.themeConceptScheme}}
             @selected={{this.themeRecords}}
             @onChange={{this.updateThemeFilter}}
             @label="Thematisch"
-            @isOpenInitially={{true}}
+            @isOpenInitially={{not (is-empty this.typeRecords)}}
```
wolfderechter:
```suggestion
            @isOpenInitially={{not (is-empty this.themeRecords)}}
```

app/controllers/search.js
```diff
@@ -32,6 +32,7 @@ export default class SearchController extends Controller {
   @tracked themes = [];
   @tracked authorities = [];
   @tracked administrativeUnits = [];
+  @tracked useDefaultUnitFilter = true;
```
wolfderechter: I would rename this to `useDefaultAdminUnitFilter` maybe

wolfderechter (approved):
Looks good, but I feel like we are missing a button to reset the filters. The queryparams can get quite sticky now, once the useDefaultUnitFilter is set to false users can't really go back to the default view. Maybe needs a followup ticket

PR #472, opened 2026-04-30
Convert vendor-selector to .gjs single-file component

cecemel:


Windvis:
I'll leave this one out for now. We can just run the codemod to convert all the components in the future.

PR #474, opened 2026-05-19
[DL-7361] Conditionally redirect to an external login page

Windvis:
In production we want to display an external login page instead of the default one.

Frontend based implementation of https://github.com/lblod/app-digitaal-loket/pull/736 which doesn't have the same issue.

### DIFF
--- a/app/routes/login.js
+++ b/app/routes/login.js
@@ -1,10 +1,16 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import config from 'frontend-loket/config/environment';
 
 export default class LoginRoute extends Route {
   @service() session;
 
   beforeModel() {
-    this.session.prohibitAuthentication('index');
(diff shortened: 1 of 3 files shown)


app/routes/login.js
```diff
@@ -1,10 +1,16 @@
...
 
 export default class LoginRoute extends Route {
   @service() session;
 
   beforeModel() {
-    this.session.prohibitAuthentication('index');
+    if (this.session.prohibitAuthentication('index')) {
+      const { loginPageRedirectUrl } = config;
+      if (!loginPageRedirectUrl.startsWith('{{')) {
+        window.location = loginPageRedirectUrl;
```
Windvis, line 12: The current setup would also redirect users who directly browse to `/login`, but it needs to be confirmed by Pieter if this is wanted or not.

If it's not we need to go with the requireAuthentication based solution again.

Edit: confirmed during standup.

PR #475, opened 2026-06-05
Prefill the "vendor selector" filter

Windvis:
Depends on https://github.com/lblod/app-digitaal-loket/pull/741

**Test instructions**
- Set up the app with the code from this pr: https://github.com/lblod/app-digitaal-loket/pull/741
- log in as a worship user (For example: St. Amatus Vleteren)
- Go to the Bedienaren and mandatenbeheer modules and check that the "bron" filter is prefilled (only works for Vanden Broele at the moment)


app/components/eredienst-mandatenbeheer/vendor-selector.gjs
```diff
@@ -0,0 +1,42 @@
...
+import { action } from '@ember/object';
+import PowerSelect from 'ember-power-select/components/power-select';
+
+const NO_PROVENANCE_VENDOR_ID = 'none';
+const VANDEN_BROELE_VENDOR_ID = 'b1e41693-639a-4f61-92a9-5b9a3e0b924e';
+
+export default class MandatenbeheerVendorSelectorComponent extends Component {
+  // TODO: hardcoded for now, only one real vendor exists in production.
+  // Once more vendors exist, fetch them from the API
+  // (store.query('vendor', { page: { size: 100 }, sort: 'name' })) instead.
```
Windvis: When testing I did notice some other (test) vendors like `iDemoVedendor#2000`. I assume we don't have any "real" vendors yet, so still kept it as is, but I can also change it if needed. We load the data already anyways.

app/routes/eredienst-mandatenbeheer/mandatarissen.js
```diff
@@ -53,10 +55,28 @@ export default class EredienstMandatenbeheerMandatarissenRoute extends Route.ext
...
+  beforeModel(transition) {
     const mandatenbeheer = this.modelFor('eredienst-mandatenbeheer');
     this.mandatenbeheer = mandatenbeheer;
+
+    const qps = transition.to.queryParams;
+
+    if (!this.hasInitializedVendorDefault) {
+      this.hasInitializedVendorDefault = true;
+
+      if (!qps.vendorId) {
```
Windvis, line 70: Not 100% sure if this is the behaviour we want. At the moment we will always set it to the vendor id if the query param isn't set (only on the first page load).

But it looks a bit strange if the user clears the filter and then refreshes for example, since it would then get reset to the default again.

app/components/eredienst-mandatenbeheer/vendor-selector.gjs
cecemel: That's QA data. Either case the hard coded ones, could be replace query the vendors that server the eenheid. So basically query the list there is in the DB. just a bit thorn by the "no vendor case" but not sure i want to tackle that now and it's worth the effort.

app/routes/eredienst-mandatenbeheer/mandatarissen.js
cecemel: Mm yeah; not ideal. But if you pass the parameters coming from a parent route, it's a complication AND a thing to consider when eventually this gets moved to a subapp. So i would keep it like that. It's a bit weird, but also, likely users don't always have very specific view on how exactly refresh behaviour with query params *should* work :-)

cecemel (changes requested):

app/components/eredienst-mandatenbeheer/vendor-selector.gjs
Windvis: With the `"no vendor case" ` you mean users who do not have a vendor, so the dropdown would only display "loket for lokale besturen" which would then not really do anything either?

We can just hide the select for those cases maybe?

app/routes/eredienst-mandatenbeheer/mandatarissen.js
Windvis: I've added an `"all"` "id" and set that when the user clears the selector, so after a reload we still know they cleared it.

app/components/eredienst-mandatenbeheer/vendor-selector.gjs
cecemel: good idea

app/routes/eredienst-mandatenbeheer/mandatarissen.js
cecemel: ok

cecemel (approved):

PR #476, opened 2026-06-12
[DL-7409] New loket flag cleanup

Windvis:
The new loket code has been deployed to production for a while now, so we can remove the flag and all the old code we no longer need.


cecemel (approved):
clicked around, no issue found. Ask QA to be throughout ;-0

PR #477, opened 2026-07-10
[DL-7363] fix download files containing special characters

DamonKennes:
[DL-7363]

Fix to allow files containing special characters in their titels (E.g. ^) to be downloaded as a zip.

TO TEST:
see https://github.com/lblod/frontend-toezicht-abb/pull/82 and https://github.com/lblod/frontend-worship-decisions/pull/47


Windvis (approved):
Thanks!

PR #482, opened 2026-08-10
[DL-7488] Appuniversum v4

Windvis:
Guide: https://github.com/appuniversum/ember-appuniversum/pull/566

We already resolved the deprecations so no real code changes needed other than to get the new dependency versions working correctly.


app/components/impersonation-menu.gjs
```diff
@@ -1,9 +1,7 @@
-import {
-  AuDropdown,
-  AuPill,
-  AuLink,
-  AuButton,
-} from '@appuniversum/ember-appuniversum';
```
Windvis, line 6: This barrel file was removed in v4 and not really documented either.

